PR #13, opened 2015-12-01
Fix typo in pom.xml

gchen:


haoch:
LGTM, thanks for the patch @gchen :-)

PR #25, opened 2015-12-12
EAGLE-14: Hdfs User command re-assembler

yonzhang:
test case for user command reassembler

user command reassembler

flexible hfds user command pattern from file and db

user command reassembler

test case for user command reassembler

user command reassembler

script for append

clean hdfsusercommand code and use eagleserviceconnector

test siddhi state persist and restore

add test case for snapshot/restore

add license header

user command reassembler

test case for user command reassembler

user command reassembler

flexible hfds user command pattern from file and db

script for append

user command reassembler

user command reassembler

clean hdfsusercommand code and use eagleserviceconnector

test siddhi state persist and restore

add test case for snapshot/restore

add license header

test for keeping user command


test.txt
```diff
@@ -0,0 +1 @@
+test for keep file for user command reassembler
```
haoch: Why keep the `test.txt` file in code base?

PR #31, opened 2015-12-15
[EAGLE-76] hide stream operation if role is not admin

zombieJ:
User don't have permission to add/update/delete metadata stream.


haoch:
LGTM.

1. Please rename title as `EAGLE-76 hide stream operation if role is not admin`
2. Commit with following message

		EAGLE-76 hide stream operation if role is not admin

		https://issues.apache.org/jira/browse/EAGLE-76

		Author: @zombieJ 
		Reviewer: @haoch

		Closes #31.

haoch:
Please resolve the JIRA ticket as well: https://issues.apache.org/jira/browse/EAGLE-76

PR #32, opened 2015-12-15
EAGLE-88 : Eagle web-service should have status url

RalphSu:
https://issues.apache.org/jira/browse/EAGLE-88

Clear the dependency to make dev able to start eagle-service with mvn tomcat7:run
Add template monitor resource which returns OK for external monitoring


haoch:
LGTM, will merge.

PR #33, opened 2015-12-16
EAGLE-94 Add shebang in eagle-lib.sh

qinzhaokun:
https://issues.apache.org/jira/browse/EAGLE-94

Add Shebang in eagle-lib.sh to fix the bug that eagle-lib.sh cannot be sourced in Mac.


haoch:
LGTM. Will Merge.

PR #36, opened 2015-12-23
EAGLE-99 policy distribution reporter

yonzhang:


anyway1021:
Edward, the apache ci was running on a problematic slave machine and gave the wrong build status, which made the none-valued judgement on your pr. Do you need to close and reopen this pr to re-fresh the status?

yonzhang:
solve Jenkins problem

yonzhang:
even after I re-open this PR, it still shows jenkins error

Thanks
Edward

anyway1021:
This is tricky. I also tested my pr for testing, and the result is, if the CI job enables pr trigger, any failure can cause your pr to be marked as Fail; and if I disable the trigger, your pr cannot be refreshed.
In this situation, while I'm doing experiment to setup the job, the CI job may bring impact to normal pr. I should have configured it to other git repo than eagle to avoid such things happen, but any github repo out of apache org is ignored by the pr builder plugin on that Jenkins server.

So, as a whole, let me turn the job's configuration to make it succeed every time and you please reopen your pr to refresh the status, so that your process won't be blocked. Make sense?

anyway1021:
It's done, please try reopen again and see if this pr can get refreshed.

PR #39, opened 2015-12-24
EAGLE-47 Ability to audit who made changes to Eagle policies

murkrishn:
Code change includes implementation of audit functionality. Documentation of the generic API design for the same is available [here](https://github.com/apache/incubator-eagle/files/71519/eagle_47_generic_audit_api_design.pdf).

### DIFF
--- a/eagle-assembly/src/main/bin/eagle-create-table.rb
+++ b/eagle-assembly/src/main/bin/eagle-create-table.rb
@@ -51,5 +51,6 @@ createEagleTable(admin, 'hbaseResourceSensitivity')
 createEagleTable(admin, 'mlmodel')
 createEagleTable(admin, 'userprofile')
 createEagleTable(admin, 'hfdsusercommandpattern')
+createEagleTable(admin, 'serviceAudit')
 
 exit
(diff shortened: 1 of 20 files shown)


yonzhang:
One comment: why do we need named audit listener? I see HBaseStorageAudit.isAuditingRequired checks if the passed service name matches "AlertDefinitionService" or "AlertDataSourceService".

murkrishn:
For the particular implementation in HBaseStorageAudit, we would be auditing the HBase operations only for the services - AlertDefinitionService and AlertDataSourceService provided the auditing option is  also enabled . Having named registering of implementation will help us to include/remove another implementation if required.

yonzhang:
understood, please go ahead to merge

anyway1021:
@yonzhang , if we're going to run the script and merge, PRs need an approval comment on the page. The approval comment should contain a single line with a string, either ":+1:", or "LGTM", we can regard these 2 string as approval marks.

Please take notice, the marks should take 1 single line with no other messages, just like below:

:+1: 

or

LGTM

Then, when you execute the merge script, it will verify and know that this PR is reviewed and approved to be merged.

yonzhang:
:+1:

PR #46, opened 2015-12-29
EAGLE-100 Provide tools to collect hadoop jmx metrics 

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-100


yonzhang:
Comments:
1. I want to apply the patch in my local machine. It fails, you will have the same issue when you apply the patch to master. You probably need collapse all your commits into one commits, then it will not have any conflict
2. the script only query jmx with "qry=Hadoop:*", then we will miss other metric like java.lang:* etc. Can the script download everything, but use the configuration in eagle-collector.conf to specify which parts to preserve.

I will continue to review

yonzhang:
more comments:
1) site attribute should be part of metric dimension
2) I would suggest we change attribute name "type" to "component"

yonzhang:
another comments: can we make this tool into eagle-external?

qingwen220:
Yes, I agree with your comments. Thanks. I will update the scripts.

PR #50, opened 2016-01-02
EAGLE-97 Enable GC Log monitoring

sunlibin:


yonzhang:
can the data model for CMS GC events be also trimmed/modified for other GC strategy for example g1?

sunlibin:
@yonzhang, Yes, the data model abstract the heap usage and gc pause time, it also works for G1 GC

RalphSu:
Looks good. And UT runs ok.

PR #53, opened 2016-01-05
EAGLE-112 Should support sum aggregation in front end for slide window function

zombieJ:
Also In front end, we should able to rename the aggregation value with a meaning name, so the downstream components like alert notification part could display more meaning full information


haoch:
LGTM

sunlibin:
Looks good to me, verified in my sandbox, the function works well

PR #55, opened 2016-01-06
EAGLE-39 Policy State Management

yonzhang:
executor state management mainly for policy state durability for failover
1. basic implementation of state snapshot, delta event persistence, snapshot recover, delta event recovery
2. by default it is disabled
3. need improve: batch insert delta events to kafka, and batch ack storm
4. need performance test further.


haoch:
It has conflict with master branch now, should we merge this part of code into master firstly?

yonzhang:
This ticket need more discussion before we merge

haoch:
Ok, thanks.

PR #56, opened 2016-01-12
Eagle 119 fixunittest

yonzhang:
Siddhi does not support external time based snapshot yet, need cancel this unit test


yonzhang:
only need include unit test fix

PR #57, opened 2016-01-12
fix unit test from siddhi

yonzhang:


yonzhang:
need normalized title

PR #58, opened 2016-01-12
EAGLE-119 fix unit test from siddhi

yonzhang:


yonzhang:
some conflict in patch, need recreate

PR #64, opened 2016-01-15
EAGLE-128 : Fix unit test failure

RalphSu:
https://issues.apache.org/jira/browse/EAGLE-128


yonzhang:
is this supposed to affect all event evaluation?

RalphSu:
i was considering add streamname into the stream definition for further handling, but revert back. This line was not reverted. Yes, this is a feature breaking for all event evaluation

PR #71, opened 2016-01-22
EAGLE-95 DAM HDFS topology fails to start, if there is invalid policy

murkrishn:
Code changes include resolution to the bug - unable to start a topology with has an invalid policy defined already by skipping SiddhiRuntime execution to start for that policy. Achieved by introducing a markdown feature every policy defined (before/during topology runtime). The markdown will dictate actions related to that policy. 

Design document is available [here](https://cwiki.apache.org/confluence/display/EAG/Policy+Markdown+Feature).


yonzhang:
It looks not efficient to do two siddhi query parsing. We should do only one siddhi query parsing and catch exception when parsing fails. If parsing fails, policy should be marked down.

senthilec566:
@yonzhang  , we will modify the code not to do two siddhi query parsing..

murkrishn:
Thanks for pointing it out @yonzhang.

To avoid the same, I will still go with determining the markdown for the policy at PolicyProcessExecutor class (methods - createPolicyEvaluator and onPolicyChanged) and pass on the flag to PolicyEvaluator (will be adding another method for onPolicyUpdate which will take in a additional parameter for markdown). This will remove the query parsing at the SiddhiPolicyEvaluator.createSiddhiRuntime as we will be modifying it to have an additional parameter to be passed in for markdown. Is this approach good to go ?

yonzhang:
@murkrishn @senthilec566 I suggest a different solution. 
1. define a new interface in PolicyEvaluator called isMarkedDown()
2. It is PolicyEvaluator implementation's responsibility to determine if the policy is marked down or not. In our case, SiddhiPolicyEvaluator and MLPolicyEvaluator should implement this above interface
3. It is PolicyEvaluator implementation's responsibility to determine if the policy should be marked down based on parsing for Siddhi case. For SiddhiPolicyEvaluator, we can catch exception from createSiddhiRuntime method, then markdown this policy by setting one field(markedDown) to be true. That can be in SiddhiPolicyEvaluator constructor method and onPolicyUpdate method
4. It is PolicyProcessExecutor's responsibility to report markdown status back to eagle service. So probably we can check PolicyEvaluator's isMarkedDown after policy creation or update, and then write back to eagle service.
Please suggest if that will do well.

murkrishn:
@yonzhang , as per your design have updated the source. Could you please review now.

yonzhang:
@murkrishn this looks good to me except 2 minor issues
1. can we save one DB call if policyevaluator is not marked down when it is created or previous and current policyevaluator are both not marked down.
2. is AlertDefinitionDAOImpl still used, if not, we need delete that instead of add new code to it.

RalphSu:
@yonzhang @murkrishn 
The AlertDefinitionDAOImpl could be replaced by PolicyDefinitionDAOImpl. The last time refactor doesn't  remove alertdefinition dao impl. I filed https://issues.apache.org/jira/browse/EAGLE-142 to do this.
For this pull request, you might go ahead and leave it there, i will clean up it.

murkrishn:
@yonzhang Modified source to make eagle service calls (to update markdown details) on specific conditions instead instead of for every policy create/update. As per @RalphSu comment, the code at AlertDefinitionDAOImpl has not been modified further.

murkrishn:
@yonzhang could you let me know your thoughts on the code

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java
```diff
@@ -83,4 +85,27 @@ public AlertDefinitionDAOImpl(EagleServiceConnector connector){
...
+		try {
+			client.create(entityList, Constants.ALERT_DEFINITION_SERVICE_ENDPOINT_NAME);
+		} catch (IOException | EagleServiceClientException exception) {
+			LOG.error("Exception in updating markdown for policy in HBase", exception.getMessage());
+		} finally {
+			try {
+				if (null != client)
+					client.close();
+			} catch (IOException exception) {
+				LOG.debug("Unable to close Eagle service client, " + exception.getMessage());
```
RalphSu: Should the exception be error always? And please make sure not only the message is logged, but also the stack, you might want to pass the exception object to LOG here.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java
```diff
@@ -83,4 +85,27 @@ public AlertDefinitionDAOImpl(EagleServiceConnector connector){
...
+
+		List<AlertDefinitionAPIEntity> entityList = new ArrayList<>();
+		entity.setMarkdownReason(null != markdownReason ? markdownReason : "");
+		entity.setMarkdownEnabled(markdownEnabled);
+		entityList.add(entity);
+
+		try {
+			client.create(entityList, Constants.ALERT_DEFINITION_SERVICE_ENDPOINT_NAME);
+		} catch (IOException | EagleServiceClientException exception) {
+			LOG.error("Exception in updating markdown for policy in HBase", exception.getMessage());
```
RalphSu: Please make sure not only the message is logged, but also the stack, you might want to pass the exception object to LOG here.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/PolicyDefinitionDAO.java
```diff
@@ -21,6 +21,7 @@
 import java.util.Map;
 
 import org.apache.eagle.alert.entity.AbstractPolicyDefinitionEntity;
+import org.apache.eagle.alert.entity.AlertDefinitionAPIEntity;
```
RalphSu: Why need to have AlertDefinitionAPIEntity in PolicyDefinitionDAO? Ideally PolicyDefinitionDAO suppose not knowing any the specific subtype of AbstractPolicyDefinitionEntity.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/PolicyDefinitionEntityDAOImpl.java
```diff
@@ -17,14 +17,17 @@
 package org.apache.eagle.policy.dao;
 
 import org.apache.eagle.alert.entity.AbstractPolicyDefinitionEntity;
+import org.apache.eagle.alert.entity.AlertDefinitionAPIEntity;
```
RalphSu: Same as above, let us make PolicyDefinitionEntityDAO/PolicyDefinitionEntityDAOImpl not knowing specific subtype of alert definition. Ideally, all entity/daos that specific to Alerts should be refactor out from eagle-policies. We haven't achieve this goal, but let us keep it clear as much as possible

RalphSu: And the generic template parameter T is introduced for the purpose. I think we don't need to direct reference to AlertDefinitionAPIEntity here.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java
murkrishn: Exception object added to the log.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java
murkrishn: Exception object added to the log.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/PolicyDefinitionDAO.java
murkrishn: Unused import AlertDefinitionAPIEntity removed.

eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/PolicyDefinitionEntityDAOImpl.java
murkrishn: The updatePolicyDetails() already uses T for getting the entity. 
The logic inside of casting to AlertDefinitionAPIEntity and adding values specific to the class has been moved outside the impl class.

yonzhang:
it is not correct for showing log stack. Normally, it is LOG.error("some message", ex) where ex is the exception object which is passed from catch clause.

murkrishn:
@yonzhang Yes, the syntax you have mentioned is used.

senthilec566:
@yonzhang  , Hi Edward , Can you take a look into code .. If it passes all checks can we close this PR ??

yonzhang:
yes, if that passed all checks, let us close this PR. But when I apply this patch and try to test, it has the following issues

LM-SJN-00876985:incubator-eagle yonzhang$ git apply 71.patch
71.patch:127: trailing whitespace.
    private SiddhiPolicyEvaluatorUtility policyEvaluatorUtility = new SiddhiPolicyEvaluatorUtility();
71.patch:143: trailing whitespace.

71.patch:145: space before tab in indent.
       	try {
71.patch:148: space before tab in indent.
          	markdownEnabled = true;
71.patch:182: trailing whitespace.
	    runtime.outputFields = null;
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: patch failed: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java:326
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java: patch does not apply
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: patch failed: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java:250
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java: patch does not apply
error: patch failed: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java:250
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java: patch does not apply
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: patch failed: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java:39
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/executor/PolicyProcessExecutor.java: patch does not apply
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory

murkrishn:
@yonzhang Edward, let me redo the PR

PR #76, opened 2016-01-27

senthilec566:
Captured design here in Confluence :

We may need to change the way to scan .. As per Edward Suggestion , We have to use  Java SPI .. For now , Notification Plugin using Reflection API to scan and load all notification plugins.


senthilec566:
Missed to add table creation scritps :( .. Will add it soon.

senthilec566:
Now code is ready for Review , the only pending thing is  which API is good for Scanning Classes , Java SPI or Reflection ?? 

Right now Notification Plugin Using Reflection tool for scanning entities.

The only problem i could see in Java SPI is 
you have to add 
META-INF/server/<your abstract class/interface>/list of classes which extends/implements { i may be wrong - correct me if its wrong }.

But in Reflection API :
Reflections reflections = new Reflections("");
Set<Class<? extends NotificationPlugin>> subTypes = reflections.getSubTypesOf(NotificationPlugin.class);

RalphSu:
Eagle already have a class scanning lib. You might want to check the EntityRepositoryScanner for the current usage. We might keep stick to one scanning library

senthilec566:
@RalphSu  , Yes , Eagle has Entity Repository Scanner.  If i want to use EntityScanner, the entity should be registered with Entity Repository..

My requirement is completely different..

Say , as a user i want to use eagle monitoring  , eagle provides end to end components ..but as user i want to forward generated alert to Other System for Visuvalization .. in this case in Eagle we provide Plugin to implement and write their OWN business logic .. During deployment , users will upload their integrated library .. At the time of Topology Start Eagle should start load all API's which implements NotificationPlugin ..  
I was thinking that Using Reflection API we could achieve this requirement :) .. 

Per @yonzhang  Edward's Suggestion , We can follow How PolicyManager Loads  (Using JAVA SPI )..

yonzhang:
@senthilec566 I have difficulty to apply the patch into current master (in my local), probably you need do rebase.

git apply 76.patch
error: eagle-core/eagle-alert/eagle-alert-base/src/main/java/org/apache/eagle/alert/common/AlertConstants.java: No such file or directory
error: patch failed: eagle-core/eagle-alert/eagle-alert-base/src/main/java/org/apache/eagle/alert/common/AlertEmailSender.java:1

senthilec566:
@yonzhang  , thanks Edward , Let me rebase it ..

PR #77, opened 2016-01-28

senthilec566:
Captured design here in Confluence :

We may need to change the way to scan .. As per Edward Suggestion , We have to use Java SPI .. For now , Notification Plugin using Reflection API to scan and load all notification plugins.

Pls refer this Pull Request for more details https://github.com/apache/incubator-eagle/pull/76


yonzhang:
Thanks for redoing pull request. I don't understand why we need bind notificationType with AlertDefinitionAPIEntity. Doesn't AlertDefinitionAPIEntity has a field notificationDef where notifications are defined. Also, one alert definition should possibly have multiple notification types associated.Do I miss anything? (I did not mention notificationDef to you before, sorry about that)

senthilec566:
@yonzhang  thanks Edward.. yes you are correct..  

"one alert definition should possibly have multiple notification types associated " ,  I missed it :( ..

I'll remove tag and add it into notificationDef  { List of Notifications } .. Another advantage is we can avoid new API :) ..

senthilec566:
I think we need to change the way of parsing Notification Definition .. Will modify this tomorrow to create policy with Multiple Notifications etc..

yonzhang:
NotificationManager and NotificationPluginLoader should be singleton object, and also we need think of that multiple thread may invoke the updateNotificationPlugins method and in turn will invoke _init.
And in notifyAlert, the inconsistency may occur for one policy. This policy is deleted by still there are some alerts associated with this policy id.

senthilec566:
@yonzhang  , can you clarify this statement  

"This policy is deleted by still there are some alerts associated with this policy id"

senthilec566:
@yonzhang  is it something , that if the policy deleted still alerts can be there .. Pls help me to understand so that i can change code ..

yonzhang:
@senthilec566 yes, when the policy is deleted, it is still possible that notification executor receive an alert for this policy because policy evaluation and notification are possibly in different physical machines. So we need handle the case when the policy id for one alert entity does not exist.

yonzhang:
In NotificationPluginLoader, what is the purpose of isPluginLoaded? Also I saw that exceptions are not correctly handled.

senthilec566:
@yonzhang  , In Notification isPluginLoaded is unused ( i was using different approach previously ), i'll remove it . As you mentioned onAlert & updataPlugin() needs improvement .. Working on it .. Will push code once it ready and passed all test cases.

senthilec566:
@yonzhang  Modified code to support more than one Notification in Policy,,  I have removed old way of deserializing notification conf and added to deserialize object to Key Values .. Pls have a look and let me your comments.. Still working on it to improve few things..

senthilec566:
Looks like Hadoop Metric Test Case failing :(

Failed tests: 
  testNameNodeLag(org.apache.eagle.hadoop.metric.TestHadoopMetricSiddhiQL)

senthilec566:
@yonzhang  , Can you have a look into code now? .. 

" Policy evaluation and notification are possibly in different physical machines "
In this case how do we Sync policy updates ??

yonzhang:
Multiple issues:
1. Config object should be injected when topology is started from driver program. In AlertNotificationExecutor prepareConfig, we should save Config object and use that when this bolt is running
2. In AlertDefinitionAPIEntity, we still can see "notificationType" as one of tags (although it does not generate any real impact, we should keep it consistent)

for " Policy evaluation and notification are possibly in different physical machines ", I did not mean we should do sync up, but we should be aware that listOfNotifications may be null  because policy is deleted
            Set<String> listOfNotifications = policyNotificationMapping.get(entity.getTags().get(Constants.POLICY_ID));

Also, how do I verify the plugin features? Do you have unit test or sample to trigger a notification plugin? Or I have to run an end-2-end program to verify plugin?

senthilec566:
1#  Will refactor config object injection code.
2 # notificationType - this field was not used anymore.. I missed to remove it from Tags..

I can create Unit Test to verify  till Loading Plugins .. To verify this complete feature we have to run end to end program .

yonzhang:
@senthilec566 can you make sure all checks are passed, I could not apply this patch to test in my local.

LM-SJN-00876985:incubator-eagle yonzhang$ git apply 77.patch
77.patch:937: space before tab in indent.
  	<dependency>
77.patch:947: space before tab in indent.
  	<dependency>
77.patch:1133: trailing whitespace.
 *  Email Notification API to send email/sms
77.patch:1253: trailing whitespace.
 * The producer is thread safe and sharing a single producer instance across threads will generally be faster than having multiple instances.
77.patch:1256: trailing whitespace.
	INSTANCE;
error: patch failed: eagle-core/eagle-alert/eagle-alert-process/src/test/java/org/apache/eagle/alert/dao/TestAlertDefinitionDAOImpl.java:18
error: eagle-core/eagle-alert/eagle-alert-process/src/test/java/org/apache/eagle/alert/dao/TestAlertDefinitionDAOImpl.java: patch does not apply
error: patch failed: eagle-core/eagle-alert/eagle-alert-process/src/test/java/org/apache/eagle/alert/policy/TestPolicyDistributionUpdater.java:27
error: eagle-core/eagle-alert/eagle-alert-process/src/test/java/org/apache/eagle/alert/policy/TestPolicyDistributionUpdater.java: patch does not apply
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: eagle-external/hadoop_jmx_collector/config.json: No such file or directory
error: eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/dao/AlertDefinitionDAOImpl.java: No such file or directory
error: patch failed: eagle-topology-assembly/pom.xml:64
error: eagle-topology-assembly/pom.xml: patch does not apply

PR #78, opened 2016-01-29
EAGLE-140 Add eagle pipeline, single kafka monitor and hadoop metric monitor scripts

haoch:
https://issues.apache.org/jira/browse/EAGLE-140


haoch:
Merged into master branch.

PR #84, opened 2016-02-09

senthilec566:
#77 

For more details:


yonzhang:
I applied the patch and run and found many issues with code structure, error handling, existing code compatibility etc, let us hold off this patch and let me try to reconstruct the code.
The main issues are:
1. Bad error handling. Exceptions are not brought from bottom to up so you have no idea of what happens if some exception is thrown
2. No support for existing code compatibility. If we use notificationexecutor, why do we keep AlertPersistExecutor? We should make sure our code is complete. That is a bad idea to keep both old and new code. 
3. Unit test is not complete, some not passed
4. Some basic java code issues for example 		List<AlertDefinitionAPIEntity> activeAlerts = new ArrayList<AlertDefinitionAPIEntity>();, but it is assigned immediately after this. Some implementation class should be better in different package with good naming convention.

PR #86, opened 2016-02-11
EAGLE-141  fix some unit testing exception which is thrown within annoymous class

murkrishn:
Updated test cases having anonymous class throwing exception.


murkrishn:
@yonzhang

PR #87, opened 2016-02-12
EAGLE-154 hadoop-metric monitoring sandbox starter script

hdendukuri:
Create a sandbox start script that makes sure, Hadoop metric monitoring runs on new sandbox.


yonzhang:
Looks good, but 2 minor issues.
1) echo $job  should be changed to echo "$job", otherwise it will show something strange because of literal character "*"
2) unit test is passed.

yonzhang:
reviewed please merge

PR #88, opened 2016-02-17
EAGLE-81 Notification plugin architecture

yonzhang:
1. plugin configured in eagle service
2. plugin base interfaces
3. plugin-example in eagle-examples/eagle-topology-example
4. email is default plugin for backward compatibility

Need to fix in the future:
1. email has to be really asynchronized. Today it is blocked on onAlert in plugin
2. test an external jar containing notification plugin
3. alert persist should be after entitydedupe or emaildedupe. Do we really support alert eaglestore to be a plugin?
4. email utilities need to be cleaned up once after we determine the right topology


yonzhang:
to be fixed:
3. alert persist should be after entitydedupe or emaildedupe. Do we really support alert eaglestore to be a  plugin?

PR #93, opened 2016-02-18
EAGLE-158 Remove org.apache.eagle.datastream.EagleTuple

haoch:
https://issues.apache.org/jira/browse/EAGLE-158

- Remove org.apache.eagle.datastream.EagleTuple 
- Replace all  org.apache.eagle.datastream.TupleN with scala.TupleN


yonzhang:
good to go! 
We should always keep code clean and no legacy code as much as possible.

PR #107, opened 2016-03-01
EAGLE-173 auto refresh when sensitivity mark updated

zombieJ:
Auto sync service to refresh the sensitivity status.


qingwen220:
LGTM

PR #108, opened 2016-03-03
EAGLE-177 Replace Legacy "dataSource" field with "application" in all metadata

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-177


RalphSu:
+1 
Unit runs ok for me. 
Be careful this would be incompatible for existing deployment if one already deployed. The one who already load the format of old version, will need
a. Drop the old metadata that have the concept of datasource
b. Recreate the metadata (for hadoop cases, re-run the sh files to create them)

haoch:
LGTM

PR #111, opened 2016-03-04
EAGLE-61 classification supports secured cluster connection & enable Ldap authentication 

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-61


RalphSu:
Looks good to me.
+1

PR #112, opened 2016-03-04
EAGLE-185 support build cache

zombieJ:
Current UI load html template & feature module using dynamic suffix which keeping getting latest data. But for built ui folder, the data should be in cache instead of updating. Which will fast the page loading speed.


qingwen220:
LGTM +1

haoch:
LGTM

PR #116, opened 2016-03-09
EAGLE-190 JBDC Metadata Storage Extension

haoch:
https://issues.apache.org/jira/browse/EAGLE-190

JDBC as metadata storage maybe more convenient and light-weight for user to use.


haoch:
Some problem with the commit

PR #118, opened 2016-03-09
EAGLE-190 JBDC Metadata Storage Extension

haoch:
## JIRA
https://issues.apache.org/jira/browse/EAGLE-190

## INCLUDES
* Support embedded Derby in local/memory: The JDBC can support derby and mysql completely, and use derby in memory by default, so that user don’t have to install any additional dependencies even mysql to get started, and also our unit test could easily pass without any tricky modifications.
* Automatic DDL Schema Manager: JdbcEntitySchemaManager could automatically create all required tables if not existing when application is started.
* Fixed JDBC reserved field name like desc, group and so on
* Force update while inserting failed due to duplicated key exception

## NOT INCLUDED

> This part will be included in later patch but no this on, for quick delivery for must-have features.

* JDBC GroupBy Aggregation
* JDBC Order Query


haoch:
Need to rename all "user" fields into "username" because "user" is invalid for Derby.

haoch:
I think the PR is ready enough to be merged into master branch, please help review @yonzhang

PR #120, opened 2016-03-10
EAGLE-194 UI show exception alert if service error

zombieJ:
When backend error occurred, we wish to see the exception.


qingwen220:
+1

Chad-eBay:
+1

PR #121, opened 2016-03-11
EAGLE-195 policy metric display with interval of 5 min or customized interval

zombieJ:
Provide daily & hourly & Every 5 minutes interval for policy metric display


haoch:
LGTM

PR #122, opened 2016-03-11
Eagle 198  integrate new Siddhi version so that thread should not die when event evaluation fails

yonzhang:
when exception happens within Siddhi processing thread, it can continue to work

### DIFF
--- a/eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/siddhi/SiddhiPolicyEvaluator.java
+++ b/eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/siddhi/SiddhiPolicyEvaluator.java
@@ -16,6 +16,8 @@
  */
 package org.apache.eagle.policy.siddhi;
 
+import com.lmax.disruptor.ExceptionHandler;
+import com.lmax.disruptor.FatalExceptionHandler;
 import com.typesafe.config.Config;
 import org.apache.eagle.alert.entity.AbstractPolicyDefinitionEntity;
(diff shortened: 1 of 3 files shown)


eagle-core/eagle-policy/eagle-policy-base/src/main/java/org/apache/eagle/policy/siddhi/SiddhiPolicyEvaluator.java
```diff
@@ -130,6 +133,7 @@ private SiddhiRuntime createSiddhiRuntime(SiddhiPolicyDefinition policyDef) {
 
         try {
             executionPlanRuntime = siddhiManager.createExecutionPlanRuntime(executionPlan);
+            executionPlanRuntime.handleExceptionWith(new SiddhiPolicyExceptionHandler());
```
haoch: My patch is to extend capability of customizing exception handler for siddhi while it uses FatalExceptionHandler by default, so here we should override our own exception handler `SiddhiPolicyExceptionHandler` and explicitly specify by `handleExceptionWith`.

haoch:
Another problem is Ci failed due to 

~~~
Tests in error: 
  testExternalTimeSlideWindowWithGroupby(org.apache.eagle.alert.state.TestSiddhiStateSnapshotAndRestore)

Tests run: 37, Failures: 0, Errors: 1, Skipped: 1
~~~

yonzhang:
now after 3.0.5, the default exception handler is not FatalExceptionHandler, do we still need SiddhiPolicyExceptionHandler?

haoch:
If so, then we don't need to customize SiddhiPolicyExceptionHandler then.

PR #129, opened 2016-03-22
EAGLE-205 Metric dashboard support multi metrics

zombieJ:
* Support multi metric display in one chart
* support more data series in one chart


qingwen220:
LGTM  +1

PR #131, opened 2016-03-23
EAGLE-207 Management page add tips

zombieJ:
Admin management page add navigation route for admin to go back home page.
Add unsaved tips if change the config


RalphSu:
+1

PR #132, opened 2016-03-25
EAGLE-208 UI metric dashboard should support order & rename

zombieJ:
UI metric dashboard should support order & rename


qingwen220:
+1

PR #136, opened 2016-04-06
Eagle-213 Updates fail for MySql

Jashchahal:
There was a check based on the exception message, which was not correct
for MySql. Changed it to more generic, comparing with "duplicate", it will handle
more broader range now. Will re-visit this when I start testing with Postgres (later).


haoch:
LGTM.

haoch:
Thanks for the contribution.

RalphSu:
+1

PR #137, opened 2016-04-06
add developer page

yonzhang:
add developer page

### DIFF
--- a/website/_config.yml
+++ b/website/_config.yml
@@ -53,6 +53,10 @@ documentations:
         url: /docs/tutorial/classification.html
       - title: Eagle LDAP Tutorial
         url: /docs/tutorial/ldap.html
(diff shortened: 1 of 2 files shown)


website/_config.yml
```diff
@@ -53,6 +53,10 @@ documentations:
         url: /docs/tutorial/classification.html
       - title: Eagle LDAP Tutorial
         url: /docs/tutorial/ldap.html
+  - category: Developer
```
haoch: Good doc, I would prefer to Change to category name to `Development` instead of `Developer`, how do you think?

yonzhang: Yes, that is ok. This is just placeholder, we can enrich the content gradually

yonzhang:
Need more thinking on document

PR #140, opened 2016-04-09
EAGLE-229 EAGLE-228 Fix incompatible alert library

haoch:
https://issues.apache.org/jira/browse/EAGLE-229
https://issues.apache.org/jira/browse/EAGLE-228


haoch:
Not completed yet, don't merge

PR #141, opened 2016-04-11
EAGLE-214 Policy edit page need auto switch application

zombieJ:
For the policy edit page open in different application will alert this error. Need auto switch application when check the policy is in another application.


qingwen220:
👍  LGTM

PR #150, opened 2016-04-13
Add Tomcat libraries and add semicolen in controller.js.

hdendukuri:
This is a patch that make you run eagle release build


hdendukuri:
Kindly do not approve this pull request. I just need this to make release build work

PR #158, opened 2016-04-20
EAGLE-273 EAGLE-269 Fix exceeding the maximum row size of 65,535 bytes in mysql and comparisons between LONG_VARCHAR not supported

haoch:
https://issues.apache.org/jira/browse/EAGLE-273
https://issues.apache.org/jira/browse/EAGLE-269

The problem is cause the limitation of maximum row size in mysql innodb engine, so need to keep the column size relatively small, but it will cause it can't store large field, it's a little tricky, so may have following possible solutions:

* Solution One: Fix mysql innodb size to resolve the limitation.
* Solution Two: Keep the field relatively small to avoid exceeding maximum row size and make sure creating table successfully, and modify field size according to actual usage (Which is the approach we take for this problem)


qingwen220:
LGTM

PR #165, opened 2016-04-25
EAGLE-275 Eagle email alert bug: $elem["dataSource"] Alert Detected

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-275

- Two versions of email template are found. Need to delete one.
- change dataSource to application


RalphSu:
+1

yonzhang:
Will <Eagle-Host> be replaced with actual eagle service host?

qingwen220:
But we don't know what the eagle service host is, which depends on the deployment.

PR #170, opened 2016-05-02
EAGLE-289: Fix $elem["dataSource"] in alert email

pkuwm:
It seems that the file eagle-core/eagle-alert/eagle-alert-process/src/main/resources/ALERT_DEFAULT.vm is deleted in dev branch, but it is still in master branch. So we still need to fix it.


yonzhang:
reviewed please merge

qingwen220:
It's not a good practice to send the two pull requests (for both EAGLE-287 and EAGLE-289) in the same pull request

qingwen220:
For pull request #170, I don't think it's a good fix (the code needs testing)

Here is my fix in the dev branch, and tested it. 
https://github.com/apache/incubator-eagle/commit/035a9f6949e3ef378e98c9fad84c3087afef5e4d

Would you like to help merge this patch to the master branch? Thanks!

PR #173, opened 2016-05-03
EAGLE-287. Make EagleStore as the default notification method.

pkuwm:
[master acf0d6c] EAGLE-287. Make EagleStore as the default notification method. Provide an optional configuration in application.conf file for disabling eagleStore as the default notification – "eagleStoreEnabled". If "eagleStoreEnabled = false", eaglestore is disabled.


qingwen220:
Basically, Adding notification plugin is a issue related to policy definition, but not to a specific topology.
 
I have another idea to handle this issue, but not adding a configuration value for a specific topology. 

That is, let the front end provide a default notification plugin EagleStore when define a policy.  How do you think?

pkuwm:
Close.

PR #174, opened 2016-05-03
EAGLE-291 : JDBC: Update transactions fail in PostgreSQL

Jashchahal:
The transactions in Postgres was failing due to insert fail transaction already in progress and we were trying to update if insert fails. Need to rollback the failed insert transaction and then continue with update transaction. Postgres is more strict as compared to MySql, so we need save point so that we can roll it back, if required.
Had to revert changes for EAGLE-270, as @haoch mentioned that we can't compare between mediumtexts, it need to be varchar.


Jashchahal:
closing this Pul request , this is causing issues, will open new one

PR #175, opened 2016-05-03
This pull request is just for MapR integration code review

DadanielZ:
-  This is my local branch for MapR integration

Solved:
-  EAGLE-282:Auditlogparser for MapR's audit log
-  EAGLE-283:Convert fid,uid in MapR's audit log to FIle/folder name, user name
-  EAGLE-284:Connect to MapR's CLDB service

Problem:
-  Can not trigger alerts for now.

Compared to Hortonworks,  MapR depends on different jars but same package names and class name.I haven't come up with a good solution for this dependency problem, so feel free to leave comments under this pull request, any advice will be appreciated.   :)

Thanks and Regards,
Da


qingwen220:
That's really a big pull request!

I don't have too much background on this new feature, and just check the code that i am familiar with. :-)

In HDFSCommandResolver.java, the command set have been changed. Do you have thorough tests on this topology after changing this resolver?

qingwen220:
Because i have a concern. As the command set are from the hdfs audit log, but not original hdfs commands. If you change the command set here, they may mismatch the parsed value from the log

PR #177, opened 2016-05-06
EAGLE-294. If a policy metadata field is not set, add null attributes to input stream for SiddhiCEP.

pkuwm:
If a policy metadata field is not set, null attributes can not be able to add into input stream for SiddhiCEP. Alerts may not be able to detect.

So we need to put null attributes into input stream.

In the sourcecode, we should get metadata map for the stream.


haoch:
Please rename `eagle-core/eagle-data-process/eagle-stream-process-api/src/main/Resource` into `eagle-core/eagle-data-process/eagle-stream-process-api/src/main/resources` (where is `resources` instead of `Resource`)

pkuwm:
Done. Please review.

pkuwm:
PULL REQUEST #196

PR #181, opened 2016-05-11
EAGLE-236. Change project version to "0.4.0-SNAPSHOT" since "v0.3.0" has been released.

pkuwm:
I replaced all "<version>0.3.0-incubating</version>" in each file with "<version>0.4.0-SNAPSHOT-incubating</version>".

### DIFF
--- a/eagle-assembly/pom.xml
+++ b/eagle-assembly/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <artifactId>eagle-parent</artifactId>
         <groupId>org.apache.eagle</groupId>
-        <version>0.3.0-incubating</version>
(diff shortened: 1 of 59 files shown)


haoch:
Thanks for the contribution 👍 , but as more common practice, it's better to use `0.4.0-incubating-SNAPSHOT` instead of `0.4.0-SNAPSHOT-incubating` where `0.4.0-incubating` is the target version and postfix "-SNAPSHOT" stands for temporary status which maybe removed after released :-)

pkuwm:
OK. Will change it right now.

pkuwm:
Done. Can someone review it? Thanks.

pkuwm:
Merged.

PR #192, opened 2016-05-18
EAGLE-300 set debug-log dumping components' log level to INFO

anyway1021:
Previously, when trying to find meaningful logs of eagle, there were too many DEBUG log lines in service log file. The redundancy ratio could be found as this: within a 1.5GB log file, only 25+MB log lines are at INFO and ERROR level. Therefore, modify log4j configuration to ignore debug-log dumping from springframework, apache.commons, and apache.torque.

Please help review this PR. Thanks.

### DIFF
--- a/eagle-webservice/src/main/resources/log4j.properties
+++ b/eagle-webservice/src/main/resources/log4j.properties
@@ -19,3 +19,8 @@ log4j.rootLogger=DEBUG, stdout
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{ISO8601} %p [%t] %c{2}[%L]: %m%n
+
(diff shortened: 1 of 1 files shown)


RalphSu:
+1
keep log neat is very useful for troubleshooting.

haoch:
Doesn't it use INFO as default level.

anyway1021:
No, I didn't put rootLogger into INFO level arbitrarily. The reason is:
1. not all DEBUG messages are useless, sometimes meaningful DEBUG msg could benefit debugging ERRORs, we can just get rid of DEBUG log dumping, only keep DEBUG message that we're interested of, such as eagle's.
2. this way of ignoring messy DEBUG msg is just to let spring and apache-commons components log above INFO level by default, separated from rootLogger's DEBUG so that we will have more flexible controls on logs of different components.

Does these make sense?

anyway1021:
@haoch , or did you mean to set rootLogger at INFO level by default, and add components' DEBUG msg as needed?

yonzhang:
merged

PR #196, opened 2016-05-24
EAGLE-294. If a policy metadata field is not set, add null attributes to input stream for SiddhiCEP.

pkuwm:
For dev branch. 
EAGLE-294. If a policy metadata field is not set, add null attributes to input stream for SiddhiCEP.


pkuwm:
Merged.

PR #199, opened 2016-05-25
EAGLE-309: Add code formatter template

RalphSu:
Add eclipse formatter.


RalphSu:
Close this one, sine the fix is merge into branch-0.4. This PR is fixed and not valid now.

PR #203, opened 2016-05-27
EAGLE-316 Feature topology should not be added into an application

zombieJ:
add `global` mark to the topology which avoid add step in application


qingwen220:
LGTM

haoch:
Please close the PR

PR #213, opened 2016-06-02
EAGLE-135 Run Eagle on Spark Streaming 

qinzhaokun:
Eagle supports Spark Streaming framework now, and an example of monitoring Cassandra audit log has been tested.


RalphSu:
Branch-0.4 already released. And we have major change in v0.5. Looks this PR deprecated to me. Can we abandon and close this one? @yonzhang @haoch

PR #221, opened 2016-06-03
EAGLE-301 modify default varchar size to 8192

anyway1021:


yonzhang:
This is good to merge

anyway1021:
this fix still has problem with DB column length, just close this PR temporarily until further fix is completed.

anyway1021:
fix the bug caused by fixing EAGLE-301 and EAGLE-319 respectively, while merging them together, it should work find as a whole.

PR #222, opened 2016-06-03
EAGLE-319 modify java to jdbc data type mapping to be compatible with derby DB

anyway1021:
modify java to jdbc data type mapping: long-varchar to varchar, to avoid comparison error for derby DB


yonzhang:
This is good to merge

anyway1021:
only change "author" in commit "876c50a"

PR #223, opened 2016-06-03
EAGLE-303. Split event fields into table in email for clearly reading.

pkuwm:
Make alert message more clearly to read in email.


yonzhang:
we should keep severity

PR #228, opened 2016-06-07
EAGLE-330 Fix Hive ql.Parser can't parser a hive query sql with keywords.

wangyum:
[EAGLE-330](https://issues.apache.org/jira/browse/EAGLE-330):Fix Hive ql.Parser can't parser a hive query sql with keywords
Updated Parser.java and TestParser.java


pkuwm:
The bug is due to an old-version hive's bug. If we use a query like this: "SELECT user, id FROM db.table", we would get exceptions. This bug was fixed in hive-1.3 and after https://issues.apache.org/jira/browse/HIVE-10731.
We are using hive-1.2 in Eagle which has this kind of bug.
The fix looks good to me. Tested on my local side. Thanks for fixing, @wangyum

yonzhang:
merged into branch-0.4

PR #237, opened 2016-06-13
EAGLE-340: refactor bin/eagle-sandbox-starter.sh to make it easier to use

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-340

Two changes 
1. use Derby as default database for sandbox starter
2. use HiveQueryLog monitoring as a sample application

### DIFF
--- a/eagle-assembly/src/main/examples/eagle-sandbox-starter.sh
+++ b/eagle-assembly/src/main/examples/eagle-sandbox-starter.sh
@@ -15,6 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


yonzhang:
going forward in Eagle 0.5, we do not need start application topology in starter, instead we do that from eagle UI.

qingwen220:
This pull request goes with the 0.4.0 document (quick-start.html), and it can help users to setup an sample application very quickly (no extra data ingestions needed)

qingwen220:
Your idea is another way. Maybe the quick-start.html needs some updates 😄

PR #243, opened 2016-06-17
EAGLE-345 Removing -SNAPSHOT from 0.4 release branch

anyway1021:
do the following 2:
1. erase "-SNAPSHOT" suffix from version of modules. 
2. add "apache-" as a prefix of finalName in eagle-assembly/pom.xml, so that final assembled tarball starts with "apache-"

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,0 +1,114 @@
+
+Release Notes - Eagle - Version v0.4.0
+
+** Highlights **
+    * JBDC Metadata Storage Extension
+    * Topology management in remote mode including start/stop/status operations
+    * Auditlogparser for MapR's audit log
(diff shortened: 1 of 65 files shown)


julianhyde:
Are you using maven-release-plugin? (Actually it's inherited from the Apache parent POM.)

It is able to update versions in all your pom.xml files automatically, removing -SNAPSHOT before the release and adding it back after the release.

anyway1021:
@julianhyde , no, the changes of this PR is not made by maven-release-plugin. I was using maven-version-plugin to erase the -SNAPSHOT suffix. 

Actually I'm researching maven-release-plugin, trying to do some experiment with it.

anyway1021:
rebase to involve new commits coming after this PR is opened.

PR #252, opened 2016-07-07
EAGLE-356 Fix Authentication problem to query resource manager web service

haoch:


haoch:
Applied into branch-0.3 and branch-0.4

PR #255, opened 2016-07-08
Fix test building problem caused by kafka version and remove unused files

wujinhu:
Fix test building problem caused by kafka version and remove unused files


qingwen220:
LGTM

PR #258, opened 2016-07-11
Eagle 365 solve issues found in release vote for 0.4.0-incubating-rc2

anyway1021:
Delete depended .jar files.


haoch:
In fact, I prefer to keep this jar in source code, but remove from built package only, how do you think?

anyway1021:
I totally agree with you, let me close this PR and have a try.

PR #260, opened 2016-07-11
EAGLE-350: Running queue metrics monitoring 

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-350


yonzhang:
Look good to go. But in next checkin, please remove datajoin configuration and also we need think of configuring HBase as storage for hadoop monitoring data.
Now seems we are not configuring storage for application, but by default we use that.

qingwen220:
Thanks!

You are right. `dataJoinPollIntervalSec` configuration in the configuration file is useless in this feature.

i will remove it.

qingwen220:
This feature must configure Hbase as the storage

PR #264, opened 2016-07-12
EAGLE-366: Create HBase tables if not exists in Eagle 0.5

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-366


qingwen220:
This is an old feature moved from Eagle-v0.4

PR #265, opened 2016-07-12
EAGLE-372: Change AlertNotificationEntity#fields type from String to List

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-372

We should avoid using String to express complex type in the entity definition.


zombieJ:
LGTM

PR #274, opened 2016-07-25
[EAGLE-392] fix 'relativePath' error of eagle-common

anyway1021:
modify eagle-core/eagle-common/pom.xml


qingwen220:
LGTM

qingwen220:
LGTM. In Eagle-common pom parent artifactId has been revised

PR #285, opened 2016-07-29
EAGLE-399 Implement consistent RESTResponse Wrapper for GET request

haoch:
https://issues.apache.org/jira/browse/EAGLE-399


yonzhang:
what is the benefit of using async response?

haoch:
The main purpose is to wrap the response of java 8 lambda function in standard type. As to Sync/Async, I think both are OK and current consideration is we could control timeout in framework level and avoid uncontrolled IO blocking.

PR #294, opened 2016-08-03
EAGLE-403 Initialize eagle jpm app

haoch:


yonzhang:
reviewed, it is one of first tries of using new app framework

PR #299, opened 2016-08-04
fix policy change listener daemon not working issue

garrettlish:


RalphSu:
Looks this PR contains commits not merged well. Could we use another pr?

PR #308, opened 2016-08-06
HBase audit monitoring with new app framework

yonzhang:
https://issues.apache.org/jira/browse/EAGLE-420


eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/impl/ApplicationManagementServiceImpl.java
```diff
@@ -63,16 +71,40 @@ public ApplicationEntity install(ApplicationOperations.InstallOperation operatio
...
+        applicationEntity.setConfiguration(appConfig);
         applicationEntity.setMode(operation.getMode());
         ApplicationContext applicationContext = new ApplicationContext(
                 applicationProviderService.getApplicationProviderByType(applicationEntity.getDescriptor().getType()).getApplication(),
                 applicationEntity,config);
         applicationContext.onInstall();
         applicationEntityService.create(applicationEntity);
+
+        // initialize web config
+        provider.installWebConfig(ConfigFactory.parseMap(appConfig));
```
haoch: Not sure the difference between `installWebConfig ` with `applicationEntity.setConfiguration(appConfig);` ?

yonzhang:
That is right, we don't need notify application of the web related configuration. Web application should do it by itself by querying ApplicationEntity, I will modify the PR

yonzhang:
merged

PR #312, opened 2016-08-09
Hbase monitoring metadata browse API in new app framework

yonzhang:
https://issues.apache.org/jira/browse/EAGLE-428


haoch:
LGTM, except:

1. MetadataDaoFactory, but we could refactor this later
2. Prefer to rename configuration key "connection" as "metadata.connection"

If we want to use json-style configuration, we should well org configuration by `{}` object group.

eagle-server/src/main/resources/application.conf
```diff
@@ -42,6 +42,7 @@
 	"metadata":{
 		"store": "org.apache.eagle.metadata.service.memory.MemoryMetadataStore"
 	},
+	"connection":"jdbc:mysql://server.eagle.apache.org:3306/eagle",
```
haoch: Could we move the `connection` into `metadata.connection`?

RalphSu: Suggest use nest {}

haoch: Yes, I meant

```
metadata{
   connection: "jdbc:mysql://server.eagle.apache.org:3306/eagle"
}
```

yonzhang:
That is obvious. Again this is to make first use case to work, I will leave this change to app framework refine.

PR #315, opened 2016-08-09
explore new packaging method for eagle using new app framework

yonzhang:
https://issues.apache.org/jira/browse/EAGLE-430


haoch:
@yonzhang it seems to fail in unit test, but never mind, i fixed in #323

PR #316, opened 2016-08-09
[EAGLE-438] Multiple Inputs for Hadoop JMX Collector Python Script

peterkim95:
https://issues.apache.org/jira/browse/EAGLE-438

### DIFF
--- a/eagle-external/hadoop_jmx_collector/config-sample.json
+++ b/eagle-external/hadoop_jmx_collector/config-sample.json
@@ -12,18 +12,35 @@
        "https":false
      }
    },
-   "input": {
-    "component": "namenode",
-    "port": "50070",
-    "https": false
-   },
+   "inputs": [
+      {
+        "component": "namenode",
+        "host": "127.0.0.1",
+        "port": "50070",
+        "https": false,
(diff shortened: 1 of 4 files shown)


eagle-external/hadoop_jmx_collector/config-sample.json
```diff
@@ -12,18 +12,35 @@
...
+      {
+        "component": "resourcemanager",
+        "host": "127.0.0.1",
+        "port": "8088",
+        "https": false,
+        "kafka_topic": "rm_jmx_metric_sandbox"
+      },
+      {
+        "component": "datanode",
+        "host": "127.0.0.1",
```
RalphSu: let's use place holder like what we do in PR312, use server.eagle.apache.org. So that developer could simply use /etc/resolv.conf when there is different endpoint he want to change (etc, hdp sandbox)

RalphSu:
otherwise, looks good to me.
+1

yonzhang:
Thanks for the contribution, Peter, it is merged in https://github.com/apache/incubator-eagle/commit/502c7e37f842eabc4021c47053dd95bdd1c7efde
Please close this PR.

PR #317, opened 2016-08-10
[EAGLE-435] make coordiantor schedule operation exclusive in distribu…

garrettlish:
…ted deployment


RalphSu:
Looks only fix the listener wise. The intension is to guarantee all the schedule.build(), not only the listener.

RalphSu:
Looks only fix the listener wise. The intension is to guarantee all the schedule.build(), not only the listener.

PR #318, opened 2016-08-10
[EAGLE-439] Introduce policy enable/disable support

garrettlish:


eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/coordinator/PolicyDefinition.java
```diff
@@ -191,6 +200,10 @@ public String toString() {
             return String.format("{type=\"%s\",value=\"%s\"",type,value);
         }
     }
+    
+    public static enum PolicyStatus {
+    	enabled, disabled
```
RalphSu: captilized enum values to follow the code style

PR #319, opened 2016-08-10
[EAGLE-439] Introduce policy enable/disable support

garrettlish:


RalphSu:
merged, pls help to close

PR #323, opened 2016-08-10
EAGLE-442 Support ApplicationProvider to extend extensible guice modules

haoch:
## Module Registry Interfaces

* ApplicationProvider
```
ApplicationProvider{ 
   void register(ModuleRegistry) 
}
```

* ModuleRegistry
```
ModuleRegistry { 
    void register(ModuleScope scope,Module module);
    default void register(Module moduel) {
        register(GlobalScope.class, module)
    }
}
```

* ModuleScope
- Modules registered in scope of `GlobalScope` will be visible globally
- Modules registered in `MySQLMetadataStore`/`MongodbMetadataStore`/`MemoryMetadataStore` will only be visible when selected MetadataStore is active (set in configuration file)

```
ModuleScope
    |- GlobalScope
    |- MetadataStore
        |- MySQLMetadataStore
        |- MongodbMetadataStore
        |- MemoryMetadataStore
```

## Example

* **Registry**: in `HBaseAuditLogAppProvider`

```
@Override
    public void register(ModuleRegistry registry) {
        registry.register(MemoryMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(InMemMetadataDaoImpl.class);
            }
        });

        registry.register(MySQLMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(JDBCSecurityMetadataDAO.class);
            }
        });
    }
```
* **Inject**: For example, `HbaseMetadataBrowseWebResource` could get current `ISecurityMetadataDAO` with `@Inject ISecurityMetadataDAO metadataDAO` as following, instead of using [`MetadataDaoFactory`](https://github.com/apache/incubator-eagle/blob/9b10f22221b60f0f6451e71ac3396b326f9565ef/eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/service/MetadataDaoFactory.java)

```
@Inject
    public HbaseMetadataBrowseWebResource(ApplicationEntityService entityService, ISecurityMetadataDAO metadataDAO){
        this.entityService = entityService;
        this.dao = metadataDAO;
    }
```

## More Use Cases
* ExampleResource
* ExampleService
* ExampleApplicationProviderTest#testApplicationExtensions

## JIRA
* https://issues.apache.org/jira/browse/EAGLE-442
* https://issues.apache.org/jira/browse/EAGLE-431


yonzhang:
I have concern with module scope, application only needs know what is current metadata store type, looks there is no need for scope. Do you have other use cases where application need know about framework.
But I think you can still merge, and we can watch

haoch:
I think ApplicationProvider is just for providing a package of static application process/metadata/extensions but not for runtime.

As following typical example, for different scope of memory/mysql, `ISecurityMetadataDAO` should bind to different implementation `InMemMetadataDaoImpl`/ `JDBCSecurityMetadataDAO`, and caller just need to inject with `@Inject ISecurityMetadataDAO dao`.

~~~
   @Override
    public void register(ModuleRegistry registry) {
        registry.register(MemoryMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(InMemMetadataDaoImpl.class);
            }
        });

        registry.register(MySQLMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(JDBCSecurityMetadataDAO.class);
            }
        });
    }
~~~

While if register with current metadata store instance, then the code have to be as following, it will bind the register interface with MetadataStore and not extensible enough.

~~~
@Override
    public void register(MetadataStore store, ModuleRegistry registry) {
       if(store instance of MemoryMetadataStore){
            registry.register(MemoryMetadataStore.class, new AbstractModule() {
               @Override
               protected void configure() {
                 bind(ISecurityMetadataDAO.class).to(InMemMetadataDaoImpl.class);
              }
        });
       } else if (store instance of MySQLMetadataStore){
        registry.register(new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(JDBCSecurityMetadataDAO.class);
            }
        });
       }
   }
~~~

I think both maybe ok, and have similar implementation code, but the first implementation should be more flexible, especially in future we may be able to extend more than MetadataStore.

PR #327, opened 2016-08-10
Eagle assembly for new app framework

yonzhang:
https://issues.apache.org/jira/browse/EAGLE-437
https://issues.apache.org/jira/browse/EAGLE-430


yonzhang:
Already reviewed by Hao in patch 315

PR #331, opened 2016-08-11
EAGLE-455 Add JDBCMetadataQueryService, JDBCDataSourceConfig, GuiceJUnitRunner

haoch:
- Enhanced JDBC Metadata Framework by adding JDBCMetadataQueryService, JDBCDataSourceConfig, GuiceJUnitRunner
- Refine `JDBCSecurityMetadataDAO` with `JDBCMetadataQueryService` and add unit test case `JDBCSecurityMetadataDAOTest`

### DIFF
--- a/eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/service/JDBCSecurityMetadataDAO.java
+++ b/eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/service/JDBCSecurityMetadataDAO.java
@@ -20,13 +20,16 @@
 package org.apache.eagle.security.service;
 
 import com.google.inject.Inject;
-import com.typesafe.config.Config;
+import org.apache.eagle.metadata.store.jdbc.JDBCMetadataQueryService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.sql.*;
-import java.util.ArrayList;
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.util.Collection;
+import java.util.Collections;
 
 /**
  * Since 8/8/16.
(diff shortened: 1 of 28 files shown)


eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/service/JDBCSecurityMetadataDAO.java
```diff
@@ -20,73 +20,72 @@
...
+
+    /**
+     * Inject datasource
+     *
+     * @param dataSource
+     */
     @Inject
-    public JDBCSecurityMetadataDAO(Config config){
-        this.config = config;
+    public JDBCSecurityMetadataDAO(DataSource dataSource, JDBCMetadataQueryService queryService) {
```
haoch: Create table if not exists when initializing.

yonzhang:
-1 for the early abstraction, we should still use existing implementation and push abstraction after we tested all the applications.

yonzhang:
@Override
    public void register(ModuleRegistry registry) {
        registry.register(MemoryMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(InMemMetadataDaoImpl.class);
            }
        });

        registry.register(JDBCMetadataStore.class, new AbstractModule() {
            @Override
            protected void configure() {
                bind(ISecurityMetadataDAO.class).to(JDBCSecurityMetadataDAO.class);
            }
        });
    }

Again this is too cumbersome, too much code for application and application developer need understand the framework to use.
Further, what if the application needs know configuration for eagle server. I still prefer to have global configuration to each application for flexibility.

haoch:
I think the implementation is to provide the foundation for make sure the code modules extensible, as how much code developer should write it's to continue simplifying the interface or DSL. I was told to think interface earlier than implementation.

> - Further, what if the application needs know configuration for eagle server

We should follow the defined interface of Application. For Application standalone mode, the configuration should be passed through configuration file, for Application managed by server, the configuration should be passed through: https://github.com/apache/incubator-eagle/blob/develop/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/ApplicationContext.java#L70

    this.config = ConfigFactory.parseMap(executionConfig).withFallback(envConfig);

For managed classes, it could easily get server config by `@Inject Config config;`

PR #335, opened 2016-08-12
EAGLE-458 Migrate eagle-jpm-spark-running using appplication framework

haoch:
https://issues.apache.org/jira/browse/EAGLE-458

### DIFF
--- a/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/ApplicationContext.java
+++ b/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/ApplicationContext.java
@@ -72,15 +72,17 @@ public class ApplicationContext implements Serializable, ApplicationLifecycle {
 
     @Override
     public void onInstall() {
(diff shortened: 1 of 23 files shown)


yonzhang:
LGTM. but to complete application conversion, we need complete configuration in metadata.xml like application.conf. Future, we may be able to reduce this duplication.

haoch:
`metadata.xml` is part of code, `application.conf` should just for standalone running mode. We should never mix them, the `value` in `metadata.xml` just stands for `defaultValue` and should be treated as configuration file. Currently we still have to manually do it. And for some application, i may have tens of configuration items like this app. 

Better solution is make the `configuration` in `metadata.xml` be part of code as designed before like:

~~~
AppConfiguration{
     @Property(name="eagle.app.configName",displayName="ConfigName", defaultValue="defaultValue")
     public String configName;
}
~~~

It should be good for:

1. Provide metadata inline in typesafe way and avoid ugly get configuration by Config.getXXX method hidden in logic code.
2. Automatically map configuration instance from typesafe Config through annotation.

haoch:
Added configuration descriptor for SparkRunningJobApp

haoch:
Merged and close.

PR #336, opened 2016-08-12
Alert State check

RalphSu:
Init draft to provide alert on state, add a state definition in policy metadata.

1.  Move input/output stream into definition
2.  Assume the intermediate stream defined in system stream definition
3.  Chain the siddhi definition, use index to identify the next handler.


RalphSu:
This is wrong pr, should to develop not master.
Closed

PR #337, opened 2016-08-12
EAGLE-464 : init draft impelmentation of multiple staged state check

RalphSu:
Init draft to provide alert on state, add a state definition in policy metadata.

Move input/output stream into definition
Assume the intermediate stream defined in system stream definition
Chain the siddhi definition, use index to identify the next handler.


yonzhang:
Good to see that stateCheck can be aligned with policy evaluation, that would be easy for user to define state rules.  LGTM.

RalphSu:
merged.

PR #344, opened 2016-08-15
EAGLE-421: JMX Metric APP use new application framework

RalphSu:
Remove the topology code since JMX app could be done in new alert engine directly.


yonzhang:
LGTM. We keep hadoop-jmx project there to put all related policies, metadata there.
But this application does not need any pre-processing logic

RalphSu:
Merged

PR #345, opened 2016-08-15
EAGLE-411 Refactor hbase related unit test cases with mockito and min…

chitin:
- Delete "zookeeper.znode.parent" and "hbase.master.ipc.address" in HbaseTestingUtility configuration.
- Support the Hbase UT which based the class to pass in meanful time

https://issues.apache.org/jira/browse/EAGLE-411


haoch:
Thanks for the contribution, could you please make sure the changes could make other test cases using `EmbeddedHbase ` able to pass?

PR #348, opened 2016-08-15
EAGLE-467: Job list apis for querying jobs regardless of the status 

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-467


wujinhu:
The serialization/deserialization of org.apache.eagle.jpm.mr.runningentity.JobConfig is a little complicated. Since it is a java HashMap,  should we consider to use java build-in  serialization/deserialization?

qingwen220:
If you declare a field as Map, then eagle will automatically adopt java build-in serialization/deserialization. 
Although JobConfig extends from Map, but eagle still regards it as a user-define type object.

haoch:
LGTM

PR #349, opened 2016-08-15
[EAGLE-465] Hive security monitoring refactor

wujinhu:


wujinhu:
remove eagle-job-common and eagle-storm-jobrunning-spout

yonzhang:
I still have concern with why Hive can't share spout with JPM applications, but we can refactor after this PR

PR #352, opened 2016-08-16
EAGLE-462: Alert Dudup

mizeng:
1. same metric generate alert should be deduped
2. user can define tags which is considered by dedup. by default will only
consider about policyId and streamId.

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/coordinator/Publishment.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/coordinator/Publishment.java
@@ -16,11 +16,11 @@
  */
 package org.apache.eagle.alert.engine.coordinator;
 
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-
-import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 /**
  * @since Apr 11, 2016
@@ -32,6 +32,7 @@ public class Publishment {
     private String type;
     private List<String> policyIds;
     private String dedupIntervalMin;
+    private List<String> dedupFields;
     private Map<String, String> properties;
     // the class name to extend the IEventSerializer interface
     private String serializer;
@@ -74,6 +75,14 @@ public class Publishment {
 
     public void setDedupIntervalMin(String dedupIntervalMin) {
         this.dedupIntervalMin = dedupIntervalMin;
+    }
+
+    public List<String> getDedupFields() {
+        return dedupFields;
+    }
+
+    public void setDedupFields(List<String> dedupFields) {
+        this.dedupFields = dedupFields;
     }
 
     public Map<String, String> getProperties() {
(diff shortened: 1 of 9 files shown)


eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/coordinator/Publishment.java
```diff
@@ -32,6 +32,7 @@
     private String type;
     private List<String> policyIds;
     private String dedupIntervalMin;
+    private List<String> customDedupFields;
```
RalphSu: dedupFields? looks custom is redundant here.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertEmailPublisher.java
```diff
@@ -69,6 +68,10 @@ public void onAlert(AlertStreamEvent event) throws Exception {
         if(event == null) {
             return;
         }
+
+        if (emailConfig.get(PublishConstants.SENDER).equals("sender@corp.com")){
```
RalphSu: what does this equals here for? For test only?

eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/coordinator/Publishment.java
mizeng: sounds better.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/resources/router/publishments.json
```diff
@@ -0,0 +1,23 @@
+[
+{
+	"name":"test-stream-output",
+	"type":"org.apache.eagle.alert.engine.publisher.impl.AlertEmailPublisher",
+	"policyIds": [
+		"perfmon_cpu_host_check", "perfmon_cpu_pool_check"
+	],
+	"properties": {
+	  "subject":"UMP Test Alert",
```
RalphSu: should be eagle test alert in "subject"

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertEmailPublisher.java
mizeng: mostly Yes.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/resources/router/publishments.json
mizeng: got it.

eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/model/AlertStreamEvent.java
```diff
@@ -35,6 +35,7 @@
     private StreamDefinition schema;
     private String createdBy;
     private long createdTime;
+    private boolean isDuplicate = false;
```
RalphSu: Looks this is only for test verification. Could we avoid add non-required method in stream event, instead use check the statistic info in deduplicator for test verify?

mizeng: currently it is only for test. But should not "duplicate" be one property for StreamEvent? It sounds more naturally.

PR #355, opened 2016-08-17
EAGLE-471: fix AlertBolt introduce un-serializble member

mizeng:


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/utils/SingletonExecutor.java
```diff
@@ -0,0 +1,32 @@
+package org.apache.eagle.alert.engine.utils;
```
RalphSu: Need apache license. Other wise this should mvn failed.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/utils/SingletonExecutor.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * @author mizeng
+ * @version v1.0
+ * @description TODO
```
RalphSu: Remove TODO/FIXME if unnecessary.

PR #359, opened 2016-08-18
EAGLE-475 - Fix generic email publisher and publish emails for absenc…

pkuwm:
…e alert.

1) Absence alerts could not be published via emails.
Root cause: in AbsencePolicyHandler, alert stream events were not emitted to alert bolt.
Solution: collect alert stream events and emit them to alert bolt.
2) Email publisher did not work.
Root cause: Email configurations were incorrect.
Solution: correct the email configurations and add email authentications.


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/runner/AlertBolt.java
```diff
@@ -111,9 +111,12 @@ public void execute(Tuple input) {
...
                 }
                 // check if specVersion is older than stream_event_version
-                else if (specVersion != null && stream_event_version != null && specVersion.contains("spec_version_") && specVersion.contains("spec_version_")){
-                    Long timestamp_of_specVersion = Long.valueOf(specVersion.split("spec_version_")[1]);
-                    Long timestamp_of_streamEventVersion = Long.valueOf(stream_event_version.split("spec_version_")[1]);
+                else if (specVersion != null && stream_event_version != null &&
+                        specVersion.contains("spec_version_") && stream_event_version.contains("spec_version_")){
+//                    Long timestamp_of_specVersion = Long.valueOf(specVersion.split("spec_version_")[1]);
+//                    Long timestamp_of_streamEventVersion = Long.valueOf(stream_event_version.split("spec_version_")[1]);
+                    Long timestamp_of_specVersion = Long.valueOf(specVersion.substring(13));
```
pkuwm: What I got for "specVersion" is "spec_version_1471478137249". split[1] would be out of array index.
So I just substring the timestamp.

yonzhang:
LGTM and merged

PR #361, opened 2016-08-18
EAGLE-473: Query error if query string contains "[]"

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-473

Fix the bug with an unit test

### DIFF
--- a/eagle-core/eagle-query/eagle-query-base/src/test/java/org/apache/eagle/query/aggregate/test/TestListQueryCompiler.java
+++ b/eagle-core/eagle-query/eagle-query-base/src/test/java/org/apache/eagle/query/aggregate/test/TestListQueryCompiler.java
@@ -30,18 +30,14 @@ public class TestListQueryCompiler {
     @Test
     public void test() throws Exception {
         try {
-            try {
-                String queryWithCondition = "JobExecutionService[@site=\"apollo\"]{*}";
-                ListQueryCompiler compiler1 = new ListQueryCompiler(queryWithCondition);
-            } catch (EagleQueryParseException ex) {}
-            try {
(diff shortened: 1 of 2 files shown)


eagle-core/eagle-query/eagle-query-base/src/test/java/org/apache/eagle/query/aggregate/test/TestListQueryCompiler.java
```diff
@@ -0,0 +1,51 @@
...
+    private final static Logger LOG = LoggerFactory.getLogger(TestListQueryCompiler.class);
+    @Test
+    public void test() throws Exception {
+        try {
+            try {
+                String queryWithCondition = "JobExecutionService[@site=\"apollo\"]{*}";
+                ListQueryCompiler compiler1 = new ListQueryCompiler(queryWithCondition);
+            } catch (EagleQueryParseException ex) {}
+            try {
+                String queryWithSquareBrackets = "JobExecutionService[@jobId=\"[A9BB756F29494F3C8281117FE58A4632/361C947C782F44748BFD840CE2FDBFB8] com.ebay.scalding.coview.CoviewsBasedOnSojornerData[site=2, runDate=2016-08-17, nDays=7}]/(4/5)\"]{*}";
```
haoch: 1) Clean sensitive info.
2) Add clear `Assert` checking, especially should equal with expected value.

qingwen220: 1) removed 
2) If the query is in a wrong format, an IllegalArgumentException will be thrown. So here Assert is to check whether there is an exception

haoch:
LGTM.

PR #369, opened 2016-08-22
[EAGLE-484] StreamRouterSpec class should override equals method

r7raul1984:
- add equals hashCode method

https://issues.apache.org/jira/browse/EAGLE-484


eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/coordination/model/PolicyWorkerQueue.java
```diff
@@ -72,6 +76,24 @@ public void setWorkers(List<WorkSlot> workers) {
         this.workers = workers;
     }
 
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
```
haoch: Should check NPE?

haoch:
LGTM.

PR #374, opened 2016-08-22
EAGLE-411 Refactor hbase related unit test cases with mockito

chitin:
EAGLE-411 Refactor hbase related unit test cases with mockito
- Mock the action of HBase reading, mock the meanful result of scanner.next(), unit test would not be enslaved to HBase.

https://issues.apache.org/jira/browse/EAGLE-411.

### DIFF
--- a/eagle-core/eagle-query/eagle-service-base/pom.xml
+++ b/eagle-core/eagle-query/eagle-service-base/pom.xml
@@ -29,10 +29,6 @@
     <artifactId>eagle-service-base</artifactId>
     <packaging>jar</packaging>
     <name>eagle-service-base</name>
-    <properties>
-        <powermock.version>1.6.5</powermock.version>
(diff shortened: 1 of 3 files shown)


baibaichen:
@haoch , I think we need write a new test class instead of modifying TestHBaseLogReader2.

eagle-core/eagle-query/eagle-service-base/pom.xml
```diff
@@ -29,6 +29,10 @@
     <artifactId>eagle-service-base</artifactId>
     <packaging>jar</packaging>
     <name>eagle-service-base</name>
+    <properties>
```
haoch: Pls. move version related properties to top-level pom.xml

chitin: I hava done it.Thanks.

PR #379, opened 2016-08-24
EAGLE-348 Alert engine base on spark streaming

r7raul1984:
- EAGLE-424 
 - Support kafka add/remove topic without requiring restart of the alert engine 
 - Read spec metadata by restful web app
 - Create UnitSparkTopologyRunner to demo spark streaming usage
 - Update eagle-alert/README.md to describe how to start spark streaming alert engine
- EAGLE-425 
 - Use Spark Accumulator to save and recover state
 - Use Spark Accumulator to support Hybird Time Window

https://issues.apache.org/jira/browse/EAGLE-348


haoch:
One of the target is to make sure alert engine could be decoupled to run on both Storm and Spark, while they should share the same metadata, I don't understand why do we need spark specific metadata like: `SpecMetadataResource`?

r7raul1984:
For now , I fetch spec config directly from metadata service, Just skip coordinator service. So I create a simple service just return spec config. In the future , I will consider to refactor coordinator service base on Spark platform.

PR #384, opened 2016-08-25
EAGLE-498 overwrite equals method for PolicyDefinition should not con…

chitin:
EAGLE-498 overwrite equals method for PolicyDefinition should not contain field description

- Delete the equals for description in method equals.
- Delete append description in method hashCode.

https://issues.apache.org/jira/browse/EAGLE-498


haoch:
LGTM.

PR #387, opened 2016-08-25
[EAGLE-500] fix process-stopping issue of mongodb unit test case

anyway1021:
https://issues.apache.org/jira/browse/EAGLE-500


haoch:
Could we just catch the `shutdown` exception and print logs instead of throw out? Exception message matching related condition looks lot very solid :-)

anyway1021:
@haoch , modified it, please take a look.

haoch:
LGTM, pls merge.

haoch:
Merged, pls. close #387.

anyway1021:
Thanks!

PR #395, opened 2016-08-26
[EAGLE-409] Automatically recognized eagle service url with http:// o…

pkuwm:
…r https://


pkuwm:
Duplicated code. Need a better package Utils for this.

haoch:
The implementation seems a little tricky, `https` physically has nothing to do with 443, thought it is always treated as default https port. :-)

pkuwm:
Right. I just simply assumed that the default port for https is 443 and see how you would deal with it. To be more accurate, the implementation should be more complicated. I'll see if we can use any library api.

pkuwm:
Simply, can we just keep the configured url and not adding any scheme "http://" or "https://"?
Say, "server.eagle.apache.org" or "http://server.eagle.apache.org".
This config will be set by users.

PR #398, opened 2016-08-29
Eagle505 oozie auditlog parser dontn`t deal with NullPointerException

chitin:
Eagle505 oozie auditlog parser dontn`t deal with NullPointerException
- add null judgment to avoid NullPointerException.

https://issues.apache.org/jira/browse/EAGLE-505


haoch:
Pls. clean the commit history firstly. 

> `TIPS`: you are recommended to create a new branch based on latest updated `develop` instead of feature branch.

PR #399, opened 2016-08-29
Eagle505 oozie auditlog parser don't deal with NullPointerException

chitin:
Eagle505 oozie auditlog parser don't deal with NullPointerException
- add null judgment to avoid NullPointerException.

https://issues.apache.org/jira/browse/EAGLE-505


yonzhang:
Thanks for the fix. We can remove the internal try/catch and only use outer try/catch to catch exception and finally it will do ack anyway

PR #401, opened 2016-08-29
[EAGLE-508] Fix NullPointerException when zookeeper is trying to shutdown in ZookeeperEmbedded.

pkuwm:


haoch:
Pls. change to title to meaningful information which will be part of commit history.

PR #402, opened 2016-08-29
Eagle-447 convert eagle-maprfs app to use new app framework

DadanielZ:
eagle-maprfs app should use new app framework to manage its lifecycle

### DIFF
--- a/eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/hdfs/MAPRFSAuditLogParser.java
+++ b/eagle-security/eagle-security-common/src/main/java/org/apache/eagle/security/hdfs/MAPRFSAuditLogParser.java
@@ -31,37 +31,41 @@ public final class MAPRFSAuditLogParser {
     public MAPRFSAuditLogParser(){
     }
 
-    public MAPRFSAuditLogObject parse(String log) throws JSONException, ParseException {
+    public MAPRFSAuditLogObject parse(String log) throws JSONException {
         JSONObject jsonObject = new JSONObject(log);
-        String timestamp = jsonObject.getJSONObject("timestamp").getString("$date");
-        String cmd = jsonObject.getString("operation");
-        String user = jsonObject.getString("uid");
-        String ip = jsonObject.getString("ipAddress");
-        String status = jsonObject.getString("status");
-        String volumeID = jsonObject.getString("volumeId");
-        String src;
-        String dst;
-        if(jsonObject.has("srcFid")){
-            src = jsonObject.getString("srcFid");
-        }else{
-            src = "null";
-        }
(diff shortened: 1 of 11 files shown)


yonzhang:
Thanks for the improvement. Why we create a new AbstractMapRFSAuditLogApplication, can't we use AbstractHdfsAuditLogApplication? They share the same code

haoch:
ApplicationProvider and XML metadata descriptor is missing.

haoch:
Pls. refer to :
* Metadata descriptor and SPI: https://github.com/apache/incubator-eagle/tree/develop/eagle-security/eagle-security-hdfs-auditlog/src/main/resources/META-INF
* ApplicationProvider: https://github.com/apache/incubator-eagle/blob/develop/eagle-security/eagle-security-hdfs-auditlog/src/main/java/org/apache/eagle/security/auditlog/HdfsAuditLogAppProvider.java
* Better to have Unit Test to make sure all set: https://github.com/apache/incubator-eagle/blob/develop/eagle-jpm/eagle-jpm-spark-running/src/test/java/SparkRunningJobAppProviderTest.java#L30

DadanielZ:
@yonzhang Their code are almost the  same except that: "AbstractHdfsAuditLogApplication" use "HdfsAuditLogParserBolt", "AbstractMapRFSAuditLogApplication" should use "MapRFSAuditLogParserBolt".

DadanielZ:
@haoch thanks for the review, will add them later.

yonzhang:
@DadanielZ that is why there is already AbstractHdfsAuditLogApplication there, and MapRApplication can extend this and implement getParserBolt method

DadanielZ:
@yonzhang, I see now. In “ AbstractHdfsAuditLogApplication.java”, there is an overrided function which use "HdfsAuditLogParserBolt" directly:
```
@Override
public StormTopology execute(Config config, StormEnvironment environment){
...
HdfsAuditLogParserBolt parserBolt = new HdfsAuditLogParserBolt();
BoltDeclarer boltDeclarer = builder.setBolt("parserBolt", parserBolt, numOfParserTasks);
...
}
```
that's why I wrote "AbstractMapRFSAuditLogApplication" and call "MapRFSAuditLogParserBolt".

I think I should refactor the code in "AbstractHdfsAuditLogApplication" and let it call "getParserBolt"  to get parserBolt, then extend this class for mapr application:
```
@Override
public StormTopology execute(Config config, StormEnvironment environment){
...
BaseRichBolt parserBolt = getParserBolt();
BoltDeclarer boltDeclarer = builder.setBolt("parserBolt", parserBolt, numOfParserTasks);
...
}
```
Will fix it.

PR #403, opened 2016-08-29
EAGLE-510: Refactor HDFS audit log parser

peterkim95:
https://issues.apache.org/jira/browse/EAGLE-510


yonzhang:
We may not change the code to use regex matching as it is too slow. We have achieved more than 10 times performance gain by using string operations than using regex. 
This is critical important as in production we have the issue with large amount of audit log and regex significantly make performance worse

peterkim95:
I see, can we can still refactor by removing the unused method parseUser

yonzhang:
sure, we can remove that method

PR #410, opened 2016-08-31
Eagle-447 convert eagle-maprfs app to use new app framework

DadanielZ:
eagle-maprfs app should use new app framework to manage its lifecycle

- in AbstractHdfsAuditLogApplication, instead of new a HdfsAuditLogParserBolt object directly, now it call "getParserBolt()" to get BaseRichBolt


haoch:
LGTM.

haoch:
@DadanielZ pls. go ahead to merge.

PR #412, opened 2016-08-31
This is for code review only: EAGLE-432 Application status monitoring

DadanielZ:
App framework needs to probe application status periodically or on demand
- this is based on dropwizard+guava, the initial delay and period setting can be set in configuration file
ApplicationEntity contains status field, and this field's value should come from physical storm applications.
- in eagle, I add "INSTALLED", "STARTING", "STARTED", "STOPPING", "STOPPED", "UNINSTALLING", "UNINSTALLED"
- physical storm topology can only be "active", "inactive" or not existed, this info can be fetched through storm ui restapi call.

### DIFF
--- a/eagle-core/eagle-app/eagle-app-base/pom.xml
+++ b/eagle-core/eagle-app/eagle-app-base/pom.xml
@@ -102,5 +102,9 @@
             <artifactId>junit</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/resource/ApplicationResource.java
+++ b/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/resource/ApplicationResource.java
@@ -18,6 +18,8 @@ package org.apache.eagle.app.resource;
 
 
 import com.google.inject.Inject;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
 import org.apache.eagle.app.service.ApplicationManagementService;
 import org.apache.eagle.app.service.ApplicationOperations;
 import org.apache.eagle.app.service.ApplicationProviderService;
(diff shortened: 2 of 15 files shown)


DadanielZ:
hi Team, I am working on application status part and here is an initial code review, just want make sure all corner cases are covered and the work flow is correct.

My idea is:
User can trigger following status for an applicaitonEntity: 
"INSTALLED"(once an application entity is created, it is always set as "INSTALLED", so there will be no "INSTALLING" status), "STARTING",  "STOPPING", "UNINSTALLING", "UNINSTALLED"

Other status including "RUNNING" and "STOPPED" should only be determined by "background task", which combine the current (stale) status in applicationEntity and status in storm restapi call response.

current question is :
1. if users triggers an "ING" status, but this topology never get a chance to be updated in storm, what should we do?
-  for "starting", retry for a certain times automatically or wait for a certain time then set status as "FAILED" or reset it to previous status?
-  for "stopping", retry or wait or reset it to previous status?

2.  "stop" function defined in ApplicaitonContext is in fact equals to "kill" a storm topology, which is confusing, as storm support "inactive" status for a "stopped" topology. May be we should add a similar function in eagle too?

Any advice will be appreciated.

Thanks and Regards,
Da

eagle-core/eagle-app/eagle-app-base/pom.xml
```diff
@@ -102,5 +102,9 @@
             <artifactId>junit</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
haoch: use `fasterxml-jackson` as default json library.

haoch:
1. Please use NimbusClient to get topology info instead of REST api, as storm ui is not must-have for storm cluster and not single-source of truth.
2. Please consider status of both `local` and `remote` mode.

Here are some code sample about how to do it which I used in some other project: https://gist.github.com/haoch/997344dae626c8858560825e7fe2356b#file-storm_topology_status-java-L31

eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/resource/ApplicationResource.java
```diff
@@ -18,24 +18,55 @@
...
+
 @Path("/apps")
 public class ApplicationResource {
+    private final static Logger LOG = LoggerFactory.getLogger(ApplicationResource.class);
+
     private final ApplicationProviderService providerService;
     private final ApplicationManagementService applicationManagementService;
     private final ApplicationEntityService entityService;
 
+    private ApplicationEntity.Status extractStatus(JSONObject response, String appUuid) {
```
haoch: No logic code should exist in web resource class (`Controller`), push down business logic into `Service` .

eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/impl/ApplicationManagementServiceImpl.java
```diff
@@ -112,14 +112,41 @@ public ApplicationEntity uninstall(ApplicationOperations.UninstallOperation oper
...
                 applicationProviderService.getApplicationProviderByType(applicationEntity.getDescriptor().getType()).getApplication(),
                 applicationEntity,config, alertMetadataService);
-        // TODO: Check status, skip stop if already STOPPED
+        // TODO: Check status
         try {
-            applicationContext.onStop();
+            ApplicationEntity.Status status = applicationEntity.getStatus();
+            switch (status){
+                case STARTING: {
+                    LOGGER.info("application is starting, please try it latter");
```
haoch: Directly thrown out exception when status is not correct.

eagle-server/src/main/java/org/apache/eagle/server/ServerApplication.java
```diff
@@ -69,5 +78,11 @@ public void run(ServerConfig configuration, Environment environment) throws Exce
 
         // context listener
         environment.servlets().addServletListeners(new CoordinatorListener());
+
+        // run application status service in background Guice.createInjector(new CommonGuiceModule(),new ApplicationGuiceModule(), new MemoryMetadataStore());
+        Injector injector = guiceBundle.getInjector();
+        ApplicationStatusUpdateService applicationStatusUpdateService = injector.getInstance(ApplicationStatusUpdateService.class);
+        Managed updateAppStatusTask = new ApplicationTask(applicationStatusUpdateService);
+        environment.lifecycle().manage(updateAppStatusTask);
```
haoch: Correct approach to run status check service as a managed task!

eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/impl/ApplicationManagementServiceImpl.java
DadanielZ: Hi @haoch, I haven't come up with a good solution for "wrong status" case.
For example, if a topology is in "starting" process (this status change is triggered by user), but somehow this   topology has never been started in storm. In such case, "status" in applicationEntity will never be updated through "background status update task".  

Possible solutions are : 
1. Add a check function which is to update status after "starting" or "stopping", and once failed it will retry until it reaches certain times or the storm toppology status become "active" or "inactive".
   -This solution add more code complexity 
2. Or I can remove the "Stopping", "starting", and only keep "installed", "Inactive", "Active".
  - This is much easier. Complex situations like mentioned before doesn't matter.


Or I can just keep status : "Active, Inactive" for topology, which is much easier and straight forward.

DadanielZ:
@haoch  Thank you for your review! :)

PR #413, opened 2016-09-01
EAGLE-519: no data alert enhancement

garrettlish:


yonzhang:
@garrettlish thanks for the PR. Can you please add more description into this PR. This PR is pretty important and we believe that people will look at this PR to see what really happened.

yonzhang:
That is fine. I found more details in JIRA
1. no data for an entire stream won't trigger gap alert (use local time & batch window) 
2. no data policies definition is too complex, make the nodata offer optional settings in stream definition 
3. bolt parallellism will cause multiple alerts in the same time window (use loopback)

PR #416, opened 2016-09-02
Update spark history job feeder config & refactor the code

qingwen220:


eagle-jpm/eagle-jpm-spark-history/src/main/resources/application.conf
```diff
@@ -45,7 +48,6 @@
     }
   },
   "storm":{
-    worker.num: 2,
```
pkuwm: Since the app framwork has set the worker number and the this conf has been removed, we need to also move the conf in SparkHistoryJobAppProvider.xml?

qingwen220: i agree with u, and the following commit is the update

yonzhang:
@qingwen220 it is a big PR, can you please add description so later on people can reference to see what really happened.

PR #419, opened 2016-09-06
[EAGLE-524] aggregation framework-job level metrics aggregation

wujinhu:


yonzhang:
@wujinhu this is pretty important PR, can you add good description so people who are interested in can understand what are those changes for.

wujinhu:
The users want to get the aggregation results of job counters group by some dimensions(by cluster/cluster&queue/cluster&queue&user and so on).For example, the Hadoop cluster administrators want to see the trend of HDFS_BYTES_READ in the last three days group by each cluster.

To meet this requirement,  I implement this by two steps. This PR is the first step. The second step is contained in https://github.com/apache/incubator-eagle/pull/430

In this PR, the history feeder generates counter metrics when it processed history jobs. When the feeder gets one task of each job, it gets the counters values and generate metric value in each minute by ( counter value / task duration * task run time in each minute). So, we get counter metrics by each job and in each minute.

PR #420, opened 2016-09-06
[EAGLE-525] refactor oozie to new application framework

joe-hj:
- Modify kafka spout schema to StringScheme
- Modify for oozie restful service
- Filter the abnormal oozie message in oozie parser

https://issues.apache.org/jira/browse/EAGLE-525


haoch:
LGTM.

BTW: could you please help add some unit test for newly added functions?

PR #429, opened 2016-09-09
[EAGLE-526] aggregation framework

wujinhu:


yonzhang:
@wujinhu can you please add sufficient description here for people to understand what aggregation framework is used for?

PR #433, opened 2016-09-12
Fix conflict

wujinhu:


yonzhang:
@wujinhu can you elaborate what are the changes for in this PR? "Fix conflict" is a very generic title for one PR :-)

PR #434, opened 2016-09-13
[EAGLE-539] add failure category and job count by state to mr feeder

wujinhu:
https://issues.apache.org/jira/browse/EAGLE-539

1. We need to classify the errors that generated by mr jobs so that we can analysis requirements like top-n errors or top-n hosts that generate errors. I implement this by using the error messages that generated by the tasks and extract the rules from them.
2. Another requirement is get job count by job status. When we parse jobs in mr history feeder, we can save job id and job status in zookeeper, and then we can flush them to eagle server.


hsaputra:
Please add more descriptions on how you are planning to fix the issue. Thx

wujinhu:
Updated the details about this issue in the jira.

PR #437, opened 2016-09-18
EAGLE-544: Enhance dedup to support extended deduplicator

garrettlish:


yonzhang:
@garrettlish can you please add good information to this PR?

PR #438, opened 2016-09-18
EAGLE-547 Resolve duplicated view path

haoch:
https://issues.apache.org/jira/browse/EAGLE-547


yonzhang:
@haoch better please to explain why view path is duplicated.

PR #451, opened 2016-09-23
EAGLE-565 Because of the "type" field when has subqueue ,response of RM REST API doesn‘t match the SchedulerWrapper object

chitin:
EAGLE-565 Because of the "type" field when has subqueue ,response of RM REST API doesn‘t match the SchedulerWrapper object
- remove the "type" field in class Queue.

https://issues.apache.org/jira/browse/EAGLE-565


eagle-jpm/eagle-hadoop-queue/src/main/java/org/apache/eagle/hadoop/queue/model/scheduler/Queue.java
```diff
@@ -24,7 +24,6 @@
 @JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Queue {
-    private String type;
```
jhsenjaliya: can you comment this field rather than removing it, also adding the explanation I provided in the EAGLE-561 ? that would be helpful.

PR #452, opened 2016-09-27
update joiner from storm to google library

garrettlish:


RalphSu:
The default deduplicator is using org.apache.storm.guava.base to implement toString(). Suggest to remove use of storm packaged library which is versioned in storm and not easy to upgrade. PR merge to use com.googe.guava.

yonzhang:
@garrettlish please add description to this PR?

garrettlish:
@yonzhang add description, thanks for reminder

PR #454, opened 2016-09-27
[EAGLE-566] add dedup fields into publishment compare consideration

garrettlish:


RalphSu:
https://issues.apache.org/jira/browse/EAGLE-566. This PR fixes the incorrect implement of equal() and hashcode(), which cause the alertPublisherBolt not able to detect config change.

Merged

PR #455, opened 2016-09-27
EAGLE-565 Because of the "type" field when has subqueue ,response of RM REST API doesn‘t match the SchedulerWrapper object

chitin:
EAGLE-565 Because of the "type" field when has subqueue ,response of RM REST API doesn‘t match the SchedulerWrapper object
- Ignore and add comments on the "type" field in class Queue, in order to workaround YARN-4785
https://issues.apache.org/jira/browse/EAGLE-565


qingwen220:
As 'type' field is not a must to have for Eagle, it is ok to remove it. I will merge this pr

jhsenjaliya:
isnt this is same PR as #451 ?

PR #456, opened 2016-09-27
EAGLE-567: Enhance Slack publish format

mizeng:
https://issues.apache.org/jira/browse/EAGLE-567

The ticket is aim to enhance the format of slack publishment of Alert. The original message was just a raw string which is not clear for user. Add more format such as color for better viewing.


RalphSu:
We could need to refactor here, not to have these check in here.

RalphSu:
https://issues.apache.org/jira/browse/EAGLE-570 for continue refactor.

PR #457, opened 2016-09-27
[EAGLE-569]: AlertPublishImpl: Concurrency : Inplace change metadata …

RalphSu:
…cause concurrent modification issue

Author : ralphsu


RalphSu:
Fix to use copy&swap way to refresh metadata changes.

PR #468, opened 2016-10-01
[EAGLE-582] add simple Basic Authentication registration to server appl…

anyway1021:
This is one of the sub-tasks of ticket EAGLE-433, to successfully register a simple Basic Authentication authenticator to the server application.

With this check-in, when the server is up, any access to rest api method that has a @Auth annotation on it's User field will require correct credentials.


eagle-server/pom.xml
```diff
@@ -129,6 +129,14 @@
...
                     <artifactId>disruptor</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-simple</artifactId>
+                </exclusion>
```
haoch: We should not include `slf4j-simple` for most cases. And dropwizard forces using logback, so the slf4j may not work at all.

eagle-server/src/main/resources/application.conf
```diff
@@ -115,4 +115,49 @@ coordinator {
...
+    # username for basic authentication, effective only when auth.mode=simple
+    username = "admin"
+
+    # password for basic authentication, effective only when auth.mode=simple
+    password = "secret"
+  }
+
+  # for ldap authentication, effective only when auth.mode=ldap
+  ldap {
+    uri = "ldaps://ldap.ebay.corp.com:636"
```
haoch: Pls. clean up sensitive information

eagle-server/src/main/resources/application.conf
```diff
@@ -115,4 +115,49 @@ coordinator {
...
+    userFilter = "ou=ziwu,dc=ebay,dc=com"
```
haoch: Clean up sensitive info.

eagle-server/src/main/resources/application.conf
```diff
@@ -115,4 +115,49 @@ coordinator {
     initDelayMillis = 1000
     delayMillis = 30000
   }
+}
+
+# ---------------------------------------------
+# Eagle Authentication Configuration
+# ---------------------------------------------
+
+auth {
```
haoch: It's better to keep reading configuration from `configuration.yml` in dropwizard stack.

pom.xml
```diff
@@ -857,6 +861,21 @@
...
+            </dependency>
+            <dependency>
+                <groupId>javax.ws.rs</groupId>
+                <artifactId>javax.ws.rs-api</artifactId>
+                <version>${javax-ws-rs.rs-api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.glassfish.jersey.core</groupId>
+                <artifactId>jersey-server</artifactId>
+                <version>${glassfish-jersey.version}</version>
```
haoch: I think jersey-server is unnecessary?

eagle-server/src/main/resources/application.conf
anyway1021: forget to erase these text before checkin, will modify it, thanks.

eagle-server/src/main/resources/application.conf
anyway1021: forget to erase these text before checkin, will modify it, thanks.

eagle-server/pom.xml
anyway1021: sure, the fragment is to exclude it

eagle-server/src/main/resources/application.conf
anyway1021: done

pom.xml
anyway1021: removed it

eagle-server/src/main/java/org/apache/eagle/server/authentication/AuthenticationRegister.java
```diff
@@ -0,0 +1,68 @@
...
+    public AuthenticationRegister(ServerConfig serverConfig, Environment environment, Class<P> principalClass) {
+        this.serverConfig = serverConfig;
+        this.environment = environment;
+        this.principalClass = principalClass;
+    }
+
+    public void register() {
+        AuthenticationMode<User> mode = AuthenticationModeIdentifier.initiate(serverConfig.getConfig(), environment).identify();
+
+        // the registration way in the following single statement is just for dropwizard 0.7.1
```
anyway1021: done

anyway1021:
re-trigger ci build for review

anyway1021:
@haoch 
@all

Would like to merge but because some new classes are added to support switching authentication without modifying code, I think it may be better to ask you to review again.

The reason of doing this is:
1. dropwizard 0.7.1 only supports @Auth annotation to mark authentication on additional User argument of REST resource methods, means if we'd like a REST path be authenticated, we should add a parameter to the method, such as "public List list(@Auth User user)". In this way, any REST resource we don't want to authenticate could only be freed by removing the parameter.

2. Currently, while registering the Authenticators, by default API, we would register a BasicAuthProvider that doesn't support switching authentication on or off. People may think, we could skip registering it when we want to switch off the authentication, however, this approach would cause 4xx http error for unable to map the coded "@Auth" parameters correctly.

3. Based on the above 2, I  create a SwitchableBasicAuthProvider to involve switching configuration, and by registering this provider on starting the server, people could configure a param "enabled=true|false" in configuration.yml, to enable|disable authentication, without having to change code, re-compile, or re-deploy.

Besides, to make other projects able to depend on class User, I moved it from eagle-server to eagle-common under package: org.apache.eagle.common.authentication, please consider if it's appropriate, thanks.

Hopefully it's not a redundant modification. 

Michael

anyway1021:
for re-trigger the reviewer ci build

anyway1021:
re-trigger ci build

eagle-server/src/main/java/org/apache/eagle/server/authentication/authenticator/LdapBasicAuthenticator.java
```diff
@@ -0,0 +1,40 @@
...
+    private AuthenticationSettings config = null;
+
+    public LdapBasicAuthenticator(AuthenticationSettings settings) {
+        super(settings, User.class);
+    }
+
+    public Optional<User> authenticate(BasicCredentials credentials) throws AuthenticationException {
+        // TODO need to implement ldap authentication logic
+        boolean pass = true;
+        if (pass) {
```
anyway1021: it's no more than a fake placeholder, will be removed by real logic soon

PR #470, opened 2016-10-02
 [EAGLE-585] : AlertEngine: create metadata validation api

RalphSu:
Author: ralphsu


RalphSu:
Fix to provide validate api in coordinator resource. There are some dependency missing cause check skipped: not able to check extension validity since extension code might only available in runtime topology.

This fix basically do the inter-reference check and policy QL validity check.

PR #480, opened 2016-10-09
[EAGLE-597] Add API to filter publishment and single policy metadata by policyId

haoch:
Alert engine publishments API only support to provide all publishments and let client-side to filter according requirements, which is NOT an scalable implementation.

As a quick solution to fix it in some way, we should provide a API to filter publishments by policy ID, so in this patch it mainly added two APIs:

* GET /metadata/policies/{policyId}
* GET /metadata/policies/{policyId}/publishments


haoch:
It's a simple patch adding a few features, and passed CI, so will merge.

PR #481, opened 2016-10-10
[EAGLE-593] Eagle Topology assembly multi-version Hadoop support & bug fix

wujinhu:
1. User may want to support different version of hadoop in eagle jpm when uses topology asembly to be a final package. For this requirement, we add profiles to eagle-topology-assembly pom.xml. In this pom, we could add different versions of Hadoop.
2. Fix StormSubmitter does not support submit another jar
In storm StormSubmitter
----------------------------------------------------------------------------------------------
private static String submittedJar = null;
private static void submitJar(Map conf, ProgressListener listener) {
if(submittedJar==null)
{ LOG.info("Jar not uploaded to master yet. Submitting jar..."); String localJar = System.getProperty("storm.jar"); submittedJar = submitJar(conf, localJar, listener); }
else
{ LOG.info("Jar already uploaded to master. Not submitting jar."); }
}
----------------------------------------------------------------------------------------------
It can not submit another jar once submittedJar is set.
So I rewrote this class.
3. Fix MR Running Job that does not set finished jobs' state

### DIFF
--- a/eagle-core/eagle-app/eagle-app-base/pom.xml
+++ b/eagle-core/eagle-app/eagle-app-base/pom.xml
@@ -64,7 +64,7 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.9.2</version>
+            <version>${thrift.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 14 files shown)


haoch:
LGTM, will merge if tests passed.

PR #483, opened 2016-10-10
Eagle-577 UI policy management

zombieJ:
For 0.5 api, policy management api has been update. UI also need fit for that.


haoch:
LGTM

PR #486, opened 2016-10-10
EAGLE-602: Exception that Spec Version [xxx] of AlertBolt is newer th…

mizeng:
Just find exception as below:
2016-10-06 22:51:41 org.apache.eagle.alert.engine.runner.AlertBolt [WARN] Spec Version [spec_version_1475817732537] of AlertBolt is newer than Stream Event Version [null]!

Seems there's an error when get meta version from stream event or initialising failed when stream event version is null.


mizeng:
@RalphSu could you help to review?

PR #489, opened 2016-10-11
EAGLE-608 UI update: metadata/policies/policy

zombieJ:
metadata/policies/policy updated from array to entity. UI also need change.


haoch:
Merged at https://github.com/apache/incubator-eagle/commit/6f9fb4eed47c7e4b4ff1b955861a4fdbf8cf296f

PR #490, opened 2016-10-11
[EAGLE-605] Fix MR feeder provider.xml and remove useless configure

wujinhu:
https://issues.apache.org/jira/browse/EAGLE-605

1. add missing configure from application.conf to provider.xml
2. remove useless configure like : pathContainsJobTrackerName/jobTrackerName/dryRun
3. Fix bug: add IsExecutable to ApplicationDesc so that web ui will know some applications can not be started, just static


haoch:
LGTM

PR #492, opened 2016-10-11
EAGLE-609: Fix nullPointerException in DefaultDeduplicator

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-609

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert-app/src/main/resources/META-INF/providers/org.apache.eagle.alert.app.AlertUnitTopologyAppProvider.xml
+++ b/eagle-core/eagle-alert-parent/eagle-alert-app/src/main/resources/META-INF/providers/org.apache.eagle.alert.app.AlertUnitTopologyAppProvider.xml
@@ -25,6 +25,12 @@
     <configuration>
        <!-- alert topology sizing parameters -->
         <property>
+            <name>topology.name</name>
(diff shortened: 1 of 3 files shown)


eagle-core/eagle-alert-parent/eagle-alert-app/src/main/resources/META-INF/providers/org.apache.eagle.alert.app.AlertUnitTopologyAppProvider.xml
```diff
@@ -25,6 +25,12 @@
     <configuration>
        <!-- alert topology sizing parameters -->
         <property>
+            <name>topology.name</name>
+            <displayName>topology.name</displayName>
+            <value>alertUnitTopology_1</value>
```
haoch: I think we should not keep current implementation only supporting hard-coded "alertUnitTopology_1", and  coordinator should be able to easily fix it to use actual topology name accordingly.

haoch: In https://github.com/apache/incubator-eagle/commit/fc2407cd53438a31b4c1072ce7d5b08578b8e43d, it will install topology metadata automatically including topology name.

PR #493, opened 2016-10-11
[EAGLE-556] Install/Update Alert Topology Metadata when start alert engine

haoch:
Currently alert engine requires additional metadata like TopologyMeta, which in fact could be automatically generated during starting alert engine.

*Changes*

* Add "ApplicationListener" in ApplicationProvider to support extensible application lifecycle management callback listener
* Implement AlertUnitTopologyAppListener to  add topology metadata when topology is running and remove topology metadata when topology is stopped


haoch:
UT passed successfully.

PR #499, opened 2016-10-13
EAGLE-615 Jsoup parse hive sql return String without line break "\n"

chitin:
EAGLE-615 Jsoup parse hive sql return String without line break "\n"
- ADD "doc.outputSettings().prettyPrint(false);", get element value via getWholeText() not text()


haoch:
LGTM, thanks for the contribution.

PR #500, opened 2016-10-13
Eagle-604 Support view/editing application configuration after installed

zombieJ:
Support view/editing application configuration after installed


qingwen220:
LGTM

PR #501, opened 2016-10-13
[EAGLE-616] AlertEngine: Reduce the connection from alert-service to ZK

mizeng:
In QA, it used to show too much connection from alert-service to zookeeper.
We only use zk for exclusive lock and spec version notify, should not have so much connection.


mizeng:
@RalphSu could you plz help to review?

PR #503, opened 2016-10-14
EAGLE-617 Scrollbar is missing when install Map Reduce History Job Monitoring

zombieJ:
add check logic in uiSrv. When dialog close, will check current existed dialog for body 'modal-open'


wujinhu:
LGTM

PR #506, opened 2016-10-14
Eagle-612 Use LOCAL when app provider is from class file and use Cluster when from jar file

zombieJ:
When jarFile is jar file and use CLUSTER by default. (end with *.jar)
When jarFile is class and use LOCAL by default. (end with *.class)


qingwen220:
LGTM

PR #514, opened 2016-10-16
[EAGLE-621] set authentication disabled by default when no 'auth' is configured

anyway1021:
Authentication settings are configured in configuration.yml file, starting with "auth" syntax. Now make the system see authentication as disabled when no auth related syntax is set.

Also, do some refactoring to make the code clearer.


eagle-server/src/main/java/org/apache/eagle/server/authentication/BasicAuthProviderBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+        boolean needsCaching = authSettings.needsCaching();
+        MAPPING.put("simple",
+                new BasicAuthProvider<>(needsCaching ? cache(simpleAuthenticator) : simpleAuthenticator, SIMPLE_MODE_REALM));
+        MAPPING.put("ldap",
+                new BasicAuthProvider<>(needsCaching ? cache(ldapAuthenticator) : ldapAuthenticator, LDAP_MODE_REALM));
+    }
+
+    public BasicAuthProvider build() {
+        if (authSettings.isEnabled()) {
+            return MAPPING.get(authSettings.getMode());
```
haoch: Please check whether mode is valid, if not MAPPING.contains(authSettings.getMode()), then throw exception.

haoch:
Very clear refactoring, LGTM.

eagle-server/src/main/java/org/apache/eagle/server/authentication/BasicAuthProviderBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final Map<String, BasicAuthProvider<User>> MAPPING = new HashMap<>();
+    private AuthenticationSettings authSettings;
+    private Environment environment;
+
+    public BasicAuthProviderBuilder(AuthenticationSettings authSettings, Environment environment) {
+        this.authSettings = authSettings;
+        this.environment = environment;
+        Authenticator<BasicCredentials, User> simpleAuthenticator = new SimpleBasicAuthenticator(authSettings.getSimple());
+        Authenticator<BasicCredentials, User> ldapAuthenticator = new LdapBasicAuthenticator(authSettings.getLdap());
+        boolean needsCaching = authSettings.needsCaching();
```
haoch: Better to register static object in static scope instead of in instance constructor, and move check `needsCaching` to `build` method.

PR #519, opened 2016-10-17
Eagle-599 improve application config

zombieJ:
* Support JSON import
* UI adjust - separate the general fields & merge env and config tab


wujinhu:
LGTM

PR #520, opened 2016-10-17
[EAGLE-628] Fix jdbc metadata bugs

wujinhu:
https://issues.apache.org/jira/browse/EAGLE-628


eagle-core/eagle-metadata/eagle-metadata-jdbc/src/main/java/org/apache/eagle/metadata/store/jdbc/service/ApplicationEntityServiceJDBCImpl.java
```diff
@@ -18,13 +18,16 @@
 package org.apache.eagle.metadata.store.jdbc.service;
 
 
+import org.apache.commons.lang.StringUtils;
```
haoch: Checkstyle failure

PR #521, opened 2016-10-17
[Minor] Set task number of each bolt in HdfsAuditLog Application

qingwen220:
If not set, the task number of all bolts are 8


wujinhu:
LGTM

PR #528, opened 2016-10-18
[EAGLE-563] migrate eagle-hadoop-queue to use application framework

anyway1021:
Migrate eagle-hadoop-queue to use application framework.


qingwen220:
LGTM

PR #538, opened 2016-10-20
EAGLE-646: AlertUnitTopology does not rebuild the scheduler state after restarting	

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-646

1. Fix a bug that alertUnitTopology does not rebuild the scheduler state after restarting
2. Fix topology assembly dependency bug

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert-app/src/main/java/org/apache/eagle/alert/app/AlertUnitTopologyAppListener.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert-app/src/main/java/org/apache/eagle/alert/app/AlertUnitTopologyAppListener.java
@@ -18,6 +18,7 @@
 package org.apache.eagle.alert.app;
 
 import org.apache.eagle.alert.coordination.model.internal.Topology;
+import org.apache.eagle.alert.coordinator.Coordinator;
 import org.apache.eagle.alert.engine.runner.UnitTopologyRunner;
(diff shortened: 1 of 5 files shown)


eagle-core/eagle-alert-parent/eagle-alert-app/src/main/java/org/apache/eagle/alert/app/AlertUnitTopologyAppListener.java
```diff
@@ -73,6 +74,7 @@ private void updateTopologyMetadata() {
             LOG.error(result.message);
             throw new IllegalStateException(result.message);
         }
+        Coordinator.startSchedule();
```
haoch: Please call `build` directly.

eagle-core/eagle-alert-parent/eagle-alert-app/src/main/java/org/apache/eagle/alert/app/AlertUnitTopologyAppListener.java
```diff
@@ -73,6 +75,11 @@ private void updateTopologyMetadata() {
             LOG.error(result.message);
             throw new IllegalStateException(result.message);
         }
+        try {
+            coordinatorResource.build();
+        } catch (Exception e) {
+            e.printStackTrace();
```
haoch: Use LOGGER instead `e.printStackTrace()`

PR #541, opened 2016-10-20
EAGLE-648: Stream router does not work properly sometimes

garrettlish:
Stream router does not work properly when multiple policies use the same input stream with the same partition definition. It is because stream router bolt output maintains a map of the route specs, but the key of the map is streamId + partition information. The monitor stream may override each other if the streamId and partition are the same.


RalphSu:
The Root cause is the stream router bolt bug miss a case that should be multiple policy references the same stream with same group (one of them is a multiple input stream policy).

PR #547, opened 2016-10-21
EAGLE-660: Stream delete mongo implementation is not working

garrettlish:
we are using name rather than streamId to delete stream, it is incorrect, change it to use streamId


RalphSu:
close this one, will create another pr for this.

RalphSu:
See https://github.com/apache/incubator-eagle/pull/548

PR #550, opened 2016-10-22
[EAGLE-662] remove jarPath/mode from entity context

wujinhu:


eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/service/ApplicationAction.java
```diff
@@ -70,13 +71,16 @@ public ApplicationAction(Application application, ApplicationEntity metadata, Co
...
+        Map<String, Object> executionConfig = new HashMap<>(metadata.getConfiguration());
         if (executionConfig == null) {
             executionConfig = Collections.emptyMap();
         }
         if (serverConfig.hasPath(MetricConfigs.METRIC_PREFIX_CONF)) {
             LOG.warn("Ignored sever config {} = {}", MetricConfigs.METRIC_PREFIX_CONF, serverConfig.getString(MetricConfigs.METRIC_PREFIX_CONF));
         }
+
+        executionConfig.put("jarPath", metadata.getJarPath());
+        executionConfig.put("mode", metadata.getMode().name());
```
haoch: It may cause the jarPath added into configuration when installating, I think you should add only in `start` action.

PR #556, opened 2016-10-24
EAGLE-670: make kafka publisher configurable and default is async

garrettlish:
Kafka send alert in sync would limit the throughput. Make this configurable, and use async by default

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertKafkaPublisher.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertKafkaPublisher.java
@@ -28,8 +28,10 @@ import java.util.concurrent.TimeUnit;
 import org.apache.eagle.alert.engine.coordinator.Publishment;
 import org.apache.eagle.alert.engine.model.AlertStreamEvent;
 import org.apache.eagle.alert.engine.publisher.PublishConstants;
+import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,6 +46,7 @@ public class AlertKafkaPublisher extends AbstractPublishPlugin {
     private KafkaProducer producer;
     private String brokerList;
     private String topic;
+    private KafkaWriteMode mode = KafkaWriteMode.async;
 
     @Override
     @SuppressWarnings("rawtypes")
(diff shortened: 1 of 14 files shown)


haoch:
@garrettlish the implementation looks a little confusing.

Please just simply use kafka configs: https://kafka.apache.org/08/documentation.html#clientconfig and another part using async kafka producer: https://github.com/apache/incubator-eagle/blob/master/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/sink/KafkaStreamSink.java#L53-L57

# kafka properties

    producer.type = async
    batch.num.messages = 3000
    queue.buffering.max.ms	= 5000
    queue.buffering.max.messages = 10000

RalphSu:
The convention here:

we will have these properties as part of publishment and need to avoid "." in key, since mongo store doesn't support "." in key, use "_" instead.

There would be many configuration for kafka, let us focus on above properties, and request.required.acks one in this PR.

haoch:
The configuration is ok. The primary concern here is the `async` implementation, kafka producer natively support `async` mode, so that you just need pass through it instead of handle `async` thread manually which will look not very clean.

haoch:
BTW: these selected config keys are what we have evaluated in real-case and impact kafka throughput mostly, instead of not requiring you guys to include additional many kafka configurations

~~~
producer.type = async
batch.num.messages = 3000
queue.buffering.max.ms  = 5000
queue.buffering.max.messages = 10000
~~~

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertKafkaPublisher.java
```diff
@@ -96,44 +103,63 @@ public void close() {
...
                     LOG.error("Alert serialize return null, ignored message! ");
-                    return null;
+                    return;
+                }
+                if (mode == KafkaWriteMode.sync) {
+                    Future<?> future = producer.send(record);
+                    future.get(MAX_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                    succeed(mode, "");
+                } else {
+                    producer.send(record, new Callback() {
```
haoch: This is not `async` mode in kafka language

haoch: Oh ... I got what you did now, but what we talk about `async` means "kafka send message with batch in async thread"

garrettlish:
Thx @haoch @RalphSu. I thought we set async for kafka producer and what we need to change is using callback rather than future wait, it is wrong, thanks for pointing out.

I have updated the code to add kafka_client_config (list of name/value map) to specify kafka producer configurations.

By default, I only set producer.type=async. 
For batch.num.messages, queue.buffering.max.ms and queue.buffering.max.messages, I think we can use kafka producer default values. 
The only difference for default value is batch.num.messages, it is 200 if not specified. Could u please share with us what is your reason to set it to 3000? 

the kafka producer properties could be defined in publish properties as follows:
    {
            "name": "***",
            "properties" : {
                    "kafka_broker": "***",
                    "topics": "***",
                    "kafka_client_config" : [
                            {
                                    "name" : "request.requrie.acks",
                                    "value": 1
                            },
                            {
                                    "name" : "producer.type",
                                    "value": "async"
                            },
                            ...
                    ]
            }
    }

haoch:
The default value is just for example :-).

And when using kafka producer `async` mode and throughput becomes extremely larger, batch size is one of the most important configurations for tuning.

garrettlish:
Yes, cool, then we can keep the current implementation and override the batch size configuration in publish properties kafka_client_config :-)

haoch:
@garrettlish thanks.

PR #562, opened 2016-10-26
Eagle-631 Policy UI refactory

zombieJ:
for https://issues.apache.org/jira/browse/EAGLE-627 , use the validate api to check policy.


wujinhu:
LGTM. Pls merge

PR #565, opened 2016-10-26
[EAGLE-685] adjust hadoop-queue application configuration parameters according to classified style

anyway1021:
application configuration has been divided into a few sections based on parameter type.
need to adjust code to fit for the change.


wujinhu:
LGTM

PR #574, opened 2016-10-27
[Eagle 695] fetching hadoop.cluster.allocatedvirtualcores instead of hadoop.cluster.totalvirtualcores

anyway1021:
Virtual Core related metrics are additional, starting from hadoop version 2.7.x. And we were originally fetching hadoop.cluster.allocatedvirtualcores. Now it seems that hadoop.cluster.totalvirtualcores is more meaningful. So change code to fetch and show it up in hadoop-queue monitoring application.


wujinhu:
LGTM

PR #578, opened 2016-10-28
[EAGLE-697] fix wrong url-concatenation for monitoring hadoop queue

anyway1021:
For hadoop queue monitoring, when endpoint url doesn't end with a slash, the final url will be composed by concatenating endpoint directly with resource path. E.g.
endpoint: https://a.b.c:1111
resource path: ws/v1/cluster/some
Then final url is: https://a.b.c:1111ws/v1/cluster/some, which make application fail.
We should add "/" to existing constants of resource paths, and compose final url with utils.
Besides, add concrete unit test to guarantee this part functions normally.


wujinhu:
Reviewed, pls merge

PR #583, opened 2016-10-29
[EAGLE-702] fix Alert Engine JdbcDatabaseHandler select sql bug

wujinhu:


wujinhu:
This fix is already in another PR which is open.

PR #588, opened 2016-10-31
[EAGLE-705] streams is null when call /rest/apps/ api

wujinhu:


eagle-core/eagle-alert-parent/eagle-alert/alert-metadata-parent/alert-metadata-service/src/main/java/org/apache/eagle/service/metadata/resource/MetadataResource.java
```diff
@@ -249,6 +249,11 @@ public OpResult removePolicy(@PathParam("policyId") String policyId) {
     @POST
     public OpResult addPublishmentsToPolicy(@PathParam("policyId") String policyId, List<String> publishmentIds) {
         OpResult result = new OpResult();
+        if (publishmentIds.size() == 0) {
```
haoch: checkNull

haoch:
LGTM

PR #589, opened 2016-11-01
Eagle-714 Application improvement

zombieJ:
* Application detail page display as `N/A`
* use 2 line of input field (jarPath)


qingwen220:
LGTM

PR #597, opened 2016-11-02
Eagle-707 Update JPM logic & ace editor support

zombieJ:
* modify jpm text content
* update grunt script for ace editor support


haoch:
LGTM

PR #602, opened 2016-11-03
EAGLE-728 Application can't load application.properties when running local mode

chitin:
EAGLE-728 Application can't load application.properties when running local mode
- Add property in server config file

https://issues.apache.org/jira/browse/EAGLE-728


haoch:
@wujinhu pls. have a look about this PR.

wujinhu:
Reviewed. We should not add application specific level configuration to eagle server configuration. I think we can hard code, and override it if user passes this configuration from web ui.

wujinhu:
Please close this, I see another PR has been opened for the same issue.

PR #607, opened 2016-11-03
[EAGLE-727] Fix TestGroupAggregateTimeSeriesClient and TestGroupAggregateClient

baibaichen:
TestGroupAggregateTimeSeriesClient and TestGroupAggregateClient can't run correctly due to without setting hbase.coprocessor.region.classes


baibaichen:
@haoch Don't merge, I will create one PR for eagle-storage-hbase with several commits.

PR #612, opened 2016-11-07
EAGLE-728 Application can't load application.properties when running local mode

chitin:
EAGLE-728 Application can't load application.properties when running local mode
- Use hard code, and overwrite it if user passes this configuration form web ui.

https://issues.apache.org/jira/browse/EAGLE-728

### DIFF
--- a/eagle-jpm/eagle-hadoop-queue/src/main/resources/META-INF/providers/org.apache.eagle.hadoop.queue.HadoopQueueRunningAppProvider.xml
+++ b/eagle-jpm/eagle-hadoop-queue/src/main/resources/META-INF/providers/org.apache.eagle.hadoop.queue.HadoopQueueRunningAppProvider.xml
@@ -47,6 +47,12 @@
       <description>interval seconds of fetching metric from resource manager</description>
       <value>10</value>
     </property>
+    <property>
+      <name>job.name.normalization.rules.key</name>
+      <displayName>Job Name Normalization Rules Key</displayName>
+      <description>User use -Dkey=job name normalization rules</description>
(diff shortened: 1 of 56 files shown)


eagle-jpm/eagle-hadoop-queue/src/main/resources/META-INF/providers/org.apache.eagle.hadoop.queue.HadoopQueueRunningAppProvider.xml
```diff
@@ -47,6 +47,12 @@
       <description>interval seconds of fetching metric from resource manager</description>
       <value>10</value>
     </property>
+    <property>
+      <name>job.name.normalization.rules.key</name>
+      <displayName>Job Name Normalization Rules Key</displayName>
+      <description>User use -Dkey=job name normalization rules</description>
+      <value>^(.*)[0-9]{4}/[0-9]{2}/[0-9]{2}/[0-9]{2}$ => $1~ ; ^(oozie:launcher):T=(.*):W=(.*):A=(.*):ID=(?:.*)$ => $1-$2-$3-$4~ ; ^(.*)([0-9]{10})$ => $1~ ; ^[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}$ => uuid~ ; ^(.*)(?:[0-9]{8}/[0-9]{2}_[0-9]{2})$ => $1~</value>
+    </property>
   </configuration>
```
wujinhu: Please remove this configuration from these Provider xml files. Since we have 'Custom' function so that advanced users can add other configurations(like this) if they want.
If user passes this custom defined configuration, it will override the predefined value.

PR #614, opened 2016-11-07
EAGLE-743: fix a bug in streamRouter

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-743


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/java/org/apache/eagle/alert/engine/router/TestStreamRouterBolt.java
```diff
@@ -0,0 +1,150 @@
+package org.apache.eagle.alert.engine.router;
```
RalphSu: need apache license header

qingwen220:
@RalphSu I have added it. Thanks

PR #620, opened 2016-11-08
EAGLE-747 Add unit test for eagle-storage-base module

chitin:
EAGLE-747 Add unit test for eagle-storage-base module
- Add unit test.

https://issues.apache.org/jira/browse/EAGLE-747


baibaichen:
@chitin how code coverage is changed?

chitin:
The module coverage is changed to 55%.

PR #627, opened 2016-11-08
[EAGLE-754] Refactor - move ignored test class to corresponding module

baibaichen:
With this commit, improving code coverage of eagle-entity-base to 51%, code coverage of eagle-query-base to 54%


baibaichen:
missing remove ignore of  TestHBaseLogReader2

PR #632, opened 2016-11-09
[EAGLE-749] Add unit test for model.

r7raul1984:
- Add unit test for model which under alert-common moudle.
 - Fix equals() hashcode() inconsistent for PartitionedEvent,StreamEvent,Publishment.
 - Fix bug for StreamColumn BooleanType.

https://issues.apache.org/jira/browse/EAGLE-749


r7raul1984:
The module coverage will change to 56%.

PR #648, opened 2016-11-13
[EAGLE-757] fix alert JdbcDatabaseHandler connection issue

wujinhu:


haoch:
LGTM, besides follow suggestions:

* Merge `alertDao` into `metadata.store`, probably by simply override system property
* Handle possible runtime exception during object mapping.

PR #658, opened 2016-11-16
[EAGLE-773] org.apache.eagle.service.generic.MetaDataResource doesn't work correctly

baibaichen:
This commit fixes this issue and also adds UT to make sure it.

However, for introducing io.dropwizard:dropwizard-testing, we need specify the following dependencies:

1 only using logback for dropwizard
2 only using org.eclipse.jetty.orbit:javax.servlet
3 only using io.dropwizard.metrics:metrics-core and io.dropwizard.metrics:metrics-jvm

All of these are done by excluding other jars

### DIFF
--- a/eagle-core/eagle-query/eagle-service-base/pom.xml
+++ b/eagle-core/eagle-query/eagle-service-base/pom.xml
@@ -50,17 +50,59 @@
             <groupId>org.apache.eagle</groupId>
             <artifactId>eagle-common</artifactId>
             <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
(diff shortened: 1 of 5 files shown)


qingwen220:
@baibaichen, could you please give us more details on this issue?

baibaichen:
@qingwen220 , the detail is in the[  EAGLE-773](https://issues.apache.org/jira/browse/EAGLE-773). In short, dropwizard uses fasterxml, and MetadataResource uses codehaus, which is conflict.

qingwen220:
Thanks. Next time, you can paste the jira ticket link when you create the pull request, so that the jira can track the update of your pull request. 😄

PR #664, opened 2016-11-17
[eagle-785] change hbase configuration to optional

puyulu:
change hbase configuration to optional


haoch:
@puyulu why so many commits history ?

puyulu:
Close for wrong commit history

PR #667, opened 2016-11-18
EAGLE-788 Add unit test for eagle-client-base module

chitin:
EAGLE-788 Add unit test for eagle-client-base module
- Add unit test for eagle-client-base module.

https://issues.apache.org/jira/browse/EAGLE-788


haoch:
LGTM

Thanks very much for the contribution.

PR #684, opened 2016-11-25
[EAGLE-800] Use InterProcessMutex to sync operation in RunningJobManager

r7raul1984:
- Use InterProcessMutex to sync operation in RunningJobManager.
 - Use siteId to generate lockpath for InterProcessMutex. 
 - Fix some checkstyle problem.  

https://issues.apache.org/jira/browse/EAGLE-800

### DIFF
--- a/eagle-jpm/eagle-jpm-util/src/main/java/org/apache/eagle/jpm/util/jobrecover/RunningJobManager.java
+++ b/eagle-jpm/eagle-jpm-util/src/main/java/org/apache/eagle/jpm/util/jobrecover/RunningJobManager.java
@@ -41,7 +41,6 @@ public class RunningJobManager implements Serializable {
     private static final String ENTITY_TAGS_KEY = "entityTags";
     private static final String APP_INFO_KEY = "appInfo";
     private static final String ZNODE_LAST_FINISH_TIME = "lastFinishTime";
-    public static final String LOCK_PATH = "/locks";
     private final InterProcessMutex lock;
 
     private CuratorFramework newCurator(String zkQuorum, int zkSessionTimeoutMs, int zkRetryTimes, int zkRetryInterval) {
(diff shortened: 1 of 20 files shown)


eagle-jpm/eagle-jpm-util/src/main/java/org/apache/eagle/jpm/util/jobrecover/RunningJobManager.java
```diff
@@ -49,12 +53,21 @@ private CuratorFramework newCurator(String zkQuorum, int zkSessionTimeoutMs, int
         );
     }
 
-    public RunningJobManager(String zkQuorum, int zkSessionTimeoutMs, int zkRetryTimes, int zkRetryInterval, String zkRoot) {
+    public RunningJobManager(String zkQuorum, int zkSessionTimeoutMs, int zkRetryTimes, int zkRetryInterval, String zkRoot, String siteId) {
```
wujinhu: I think pass the lock path to utility class directly will be better since utility class does not need to know how to make the lock path.
How to define the lock path is decided by applications

r7raul1984: OK. I will generate lockpath in utility class.Then RunningJobManager will use it. Lockpath format is /apps/{app}/{site}/locks.

eagle-jpm/eagle-jpm-util/src/main/java/org/apache/eagle/jpm/util/jobrecover/RunningJobManager.java
```diff
@@ -49,12 +52,14 @@ private CuratorFramework newCurator(String zkQuorum, int zkSessionTimeoutMs, int
         );
     }
 
-    public RunningJobManager(String zkQuorum, int zkSessionTimeoutMs, int zkRetryTimes, int zkRetryInterval, String zkRoot) {
+    public RunningJobManager(String zkQuorum, int zkSessionTimeoutMs, int zkRetryTimes, int zkRetryInterval, String zkRoot, String lockPath) {
         this.zkRoot = zkRoot;
+        curator = newCurator(zkQuorum, zkSessionTimeoutMs, zkRetryTimes, zkRetryInterval);
+        curator.start();
```
wujinhu: Do we need catch Exception that curator could not start more than once?

r7raul1984: OK.For now,I will catch it.

r7raul1984: In the future,we should follow http://stackoverflow.com/questions/20384869/when-curatorframeworks-methods-start-and-close-shoud-be-called

PR #687, opened 2016-11-28
EAGLE-794: enable publish bolt parallelism

garrettlish:
Currently the publish is using shuffle grouping, we cannot enable parallelism for publish since we may have local cache which is unavailable across the storm cluster. We are going to use field group function to partition streams for publish bolt parallelism.


RalphSu:
Looks good to me.
+1

PR #690, opened 2016-11-28
[EAGLE-804] make interface StreamContext more abstract

r7raul1984:
[EAGLE-804] make interface StreamContext more abstract 
- Add interface StreamCounter to abstract count 

 https://issues.apache.org/jira/browse/EAGLE-804


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/evaluator/impl/SiddhiPolicyHandler.java
```diff
@@ -88,18 +88,18 @@ public void prepare(final Collector<AlertStreamEvent> collector, PolicyHandlerCo
     }
 
     public void send(StreamEvent event) throws Exception {
-        context.getPolicyCounter().scope(String.format("%s.%s", this.context.getPolicyDefinition().getName(), "receive_count")).incr();
+        context.getPolicyCounter().incr(String.format("%s.%s", this.context.getPolicyDefinition().getName(), "receive_count"));
```
wujinhu: I'm not quite sure about the performance of the incr() since there are a lot of method call send(StreamEvent event). Do you evaluate it?

PR #698, opened 2016-11-29
[EAGLE-814] fix alertId is null in AlertStreamEvent bug

wujinhu:


garrettlish:
+ 1

The issue is introduced by https://github.com/apache/incubator-eagle/commit/38ec7fc23d06c77b6e083ad2dfb9c61e729901e0#diff-2bb6be6fb3afdc59db75e1919ad89beaL97 which changed the alertId from AlertPublishEvent initialisation with random UUID to use AlertStreamEvent.alertId. However, AlertStreamEvent.alertId is not always initialised. The fix looks good.

PR #699, opened 2016-11-30
[MINOR] Add makeSSS and makeSRS in RouteSpec

r7raul1984:
- Add makeSSS and makeSRS in RouteSpec


qingwen220:
@r7raul1984 do you have any description about this change?

r7raul1984:
This refactor has the following advantages:
-   Reuseability
Other object which have RouteSpec instance can reuse the makeSSS and makeSRS logic.

-   Readability
I think the logic is domain logic. So the logic that should be in a domain object.

PR #706, opened 2016-12-01
[MINOR] Sleep 1 ms after versionId was generated

chitin:
[MINOR] Sleep 1 ms after versionId was generated
- VersionId is the same with one generated in last loop randomly, the test failed in this case.


eagle-core/eagle-alert-parent/eagle-alert/alert-metadata-parent/alert-metadata/src/test/java/org/apache/eagle/alert/metadata/impl/JdbcImplTest.java
```diff
@@ -184,6 +184,11 @@ public void test_clearScheduleState() {
             if (i >= 10 - maxCapacity) {
                 reservedOnes.add(versionId);
             }
+            try {
+                Thread.sleep(1);
+            } catch (Exception e) {
```
wujinhu: What does this used for in ut?

chitin: Make sure all versionIds genereted in all loops are different value.

wujinhu:
LGTM

PR #707, opened 2016-12-02
[MINOR] avoid adding duplicated publishment in spec

garrettlish:
Although duplicated publisher won't impact anything since we convert it to map and publisher name is the key, avoid duplicated publisher in schedule state is meaningful.


wujinhu:
LGTM, will merge

PR #708, opened 2016-12-02
[MINOR] support group by in siddhi  pattern match

wujinhu:


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/interpreter/PolicyExecutionPlan.java
```diff
@@ -32,6 +32,11 @@
     private Map<String, List<StreamColumn>> inputStreams;
 
     /**
+     * Actual input streams alias.
+     */
+    private Map<String, String> inputStreamAlias;
```
haoch: Why need to keep `inputStreamAlias` in `PolicyExecutionPlan`, isn't it only used as local variable in `PolicyExecutionPlannerImpl`?

wujinhu: IMO, PolicyExecutionPlan is the data interface for user to get some details about the execution plan. Besides, I think inputStreamAlias and inputStreams are the same level information.

wujinhu: IMO, PolicyExecutionPlan is the data interface for user to get some details about the execution plan. Besides, I think inputStreamAlias and inputStreams are the same level information.

PR #716, opened 2016-12-06
[MINOR]Fix initial sql syntax problem

chitin:
[MINOR]Fix initial sql syntax problem
- Fix sql comment.
- Modify table name to lowercase, or got error "ERROR [2016-12-06 08:16:15,738] org.apache.eagle.alert.metadata.impl.JdbcMetadataHandler: Table 'eagle.kafka_tuple_metadata' doesn't exist".


haoch:
LGTM

PR #719, opened 2016-12-07
EAGLE-830 update policy display info

zombieJ:
* Policy detail support input stream detail
* Policy list display input stream


haoch:
LGTM

PR #730, opened 2016-12-12
EAGLE-827: coordinator schedule timeout due to last fail

garrettlish:
Exclusive executor exception could cause take leadership execution exit with exception, it will cause follow-up take leadership fails.


wujinhu:
LGTM

PR #732, opened 2016-12-12
[EAGLE-837] Stream definition change does not reflect in AlertBolt

garrettlish:
Stream definition change only trigger router bolt & publisher update, we don't update corresponding alert bolt stream definition references. It will cause alert bolt still use old stream definition references, it could produce array index out of bound exception.

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/java/org/apache/eagle/alert/engine/router/TestAlertBolt.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/java/org/apache/eagle/alert/engine/router/TestAlertBolt.java
@@ -46,6 +46,8 @@ import org.apache.eagle.common.DateTimeUtil;
 import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/test/java/org/apache/eagle/alert/engine/router/TestAlertBolt.java
```diff
@@ -443,6 +447,83 @@ public void reportError(Throwable error) {
...
+        sdTest.setDescription("update the stream");
+        bolt.onAlertBoltSpecChange(boltSpecs, sds);
+        
+        LOG.info("No any change");
+        sds = new HashMap();
+        sdTest = new StreamDefinition();
+        sdTest.setStreamId(TEST_STREAM);
+        sds.put(sdTest.getStreamId(), sdTest);
+        sdTest.setDescription("update the stream");
+        bolt.onAlertBoltSpecChange(boltSpecs, sds);
```
wujinhu: It seems some tests have no asserts, does it ok?

garrettlish: asserts updated in https://github.com/apache/incubator-eagle/pull/738

PR #738, opened 2016-12-14
[EAGLE-837] fix duplicated policy update and add tests

garrettlish:
1. fix duplicated policy populated during alertBolt reload
2. refine alert bolt reload test


RalphSu:
+1

Introduced bug fixed.

PR #744, opened 2016-12-15
[EAGLE-840] add task attempt stream for bad node detection

wujinhu:


haoch:
`string:empty()` should return `bool` instead of `int`

PR #751, opened 2016-12-19
[EAGLE-844] Fix a potential NPE

lifove:
https://issues.apache.org/jira/browse/EAGLE-844


eagle-core/eagle-query/eagle-storage-hbase/src/main/java/org/apache/eagle/storage/hbase/HBaseEntitySchemaManager.java
```diff
@@ -65,7 +65,7 @@ public void init() {
         try {
             admin = new HBaseAdmin(conf);
             Map<String, EntityDefinition> entityServiceMap = EntityDefinitionManager.entities();
-            if (entityServiceMap != null || entityServiceMap.values() != null) {
+            if (entityServiceMap != null && entityServiceMap.values() != null) {
```
RalphSu: do we still need to check "entityServiceMap.values() != null"?  it should not return null

lifove: Right. It shouldn't. will update soon. Thanks~

wujinhu:
LGTM, will merge.

PR #758, opened 2016-12-28
[EAGLE-687] updating SiteEntityToRelation to accept null values

jhsenjaliya:
Currently if you create site with any of the field set to NULL, it will throw 
explored the way to use preparedStatements with explicitly specified parameters but that would require quite a large refactoring on entire metadata module.
so came up with the fix that is based on current assumptions in SiteEntityServiceJDBCImpl which is forming the sql in particular way that suits in SiteEntityToRelation.accept method.


haoch:
@r7raul1984 could you please help review this PR?

r7raul1984:
@haoch LGTM

haoch:
Thanks @jhsenjaliya and @r7raul1984 , merged now.

PR #759, opened 2016-12-28
[Minor] modify eagle-dev/merge-eagle-pr.py, changing urls from incubator-eagle…

anyway1021:
modify eagle-dev/merge-eagle-pr.py, changing urls from incubator-eagle to eagle


haoch:
LGTM

PR #765, opened 2017-01-09
[EAGLE-851]  Add validation for policy name

r7raul1984:
- Add checkPolicyName method to check the length of policy name.

https://issues.apache.org/jira/browse/EAGLE-851

### DIFF
--- a/eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
+++ b/eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
@@ -304,6 +304,15 @@
 			return false;
 		};
 
+
+		$scope.checkPolicyName = function () {
+			if($scope.policy.name.length > 50) {
+				return "length should less than 50";
+			}
+			return false;
+		};
+
+
 		$scope.addPublisherConfirm = function () {
 			if($scope.addPublisherType === "exist") {
 				$scope.publisher = $.extend({
@@ -325,6 +334,7 @@
 			return (
 				!$scope.saveLock &&
 				$scope.policy.name &&
+				!$scope.checkPolicyName() &&
 				common.number.parse($scope.policy.parallelismHint) > 0 &&
 				$scope.policy.definition.value &&
 				$scope.policy.outputStreams.length &&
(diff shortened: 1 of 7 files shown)


eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
```diff
@@ -304,6 +304,15 @@
 			return false;
 		};
 
+
+		$scope.checkPolicyName = function () {
+			if($scope.policy.name.length > 50) {
+				return "length should less than 50";
```
jhsenjaliya: should we say "policy name length should be less than 50" ?  just length sounds very generic.

eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
```diff
@@ -304,6 +304,15 @@
...
+			}
+			return false;
```
jhsenjaliya: I think it should return true since the name is in valid range at this point.

eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
```diff
@@ -325,6 +334,7 @@
 			return (
 				!$scope.saveLock &&
 				$scope.policy.name &&
+				!$scope.checkPolicyName() &&
```
jhsenjaliya: if its changed to return true, should remove !

eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
r7raul1984: This message will combine with "Name". Finally, user will see the message "Name length should less than 50".

eagle-server/src/main/webapp/app/dev/public/js/ctrls/alertEditCtrl.js
r7raul1984: Yes,I agree with you.But I see checkPublisherName return false. So I consistent with checkPublisherName  method.

jhsenjaliya:
+1, LGTM

haoch:
I think the feature should be added into backend `org.apache.eagle.service.metadata.resource.MetadataResource#validatePolicy()`

haoch:
But it should be better to add validate in both front-end and backend at same time :-)

baibaichen:
+1 on @haoch.  By the way, from the perspective of priority, backend validation would be solved at first.

PR #767, opened 2017-01-09
EAGLE-852 Application start button should be disabled when it's running, and the same for stop button.

chitin:
EAGLE-852 Application start button should be disabled when it's running, and the same for stop button.
- Add disabled judgment logic.

executable | status  | start button | stop button
---|--- | --- | ---
Y | INITIALIZED | Y | N
Y | STARTING | N | N
Y | RUNNING | N | Y
Y | STOPPING | N | N
Y | STOPPED | N | N
N | INITIALIZED | N | N

https://issues.apache.org/jira/browse/EAGLE-852

### DIFF
--- a/eagle-hadoop-metric/pom.xml
+++ b/eagle-hadoop-metric/pom.xml
@@ -45,6 +45,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


haoch:
LGTM

PR #769, opened 2017-01-10
[MINOR]  Change AlertEagleStorePlugin in sql file

r7raul1984:
- Change to org.apache.eagle.alert.engine.publisher.impl.AlertEagleStorePlugin


qingwen220:
@r7raul1984  We have two implementations of the AlertEagleStorePlugin. 
{code}
org.apache.eagle.alert.app.AlertEagleStorePlugin
org.apache.eagle.alert.engine.publisher.impl.AlertEagleStorePlugin
{code}

The first one is based on HBase, and the second one is on JDBC. As the frontend codes have bound to the second. There maybe some problem if we make the change

r7raul1984:
@qingwen220 I suggest we should rename these two classname to AlertHbaseStorePlugin and AlertJDBCStorePlugin. For now,these name is confusing.

r7raul1984:
@qingwen220 I think AlertHbaseStorePlugin  and AlertJDBCStorePlugin should be insert into the table at start time. User can choose jdbc and hbase from web UI. Also,user can config it from web UI.

qingwen220:
I think the change you suggested it OK

baibaichen:
@r7raul1984  could you please push again to make chekc ok?

eagle-server-assembly/src/main/bin/metadata-ddl.sql
```diff
@@ -134,4 +134,5 @@ INSERT INTO publishment_type(id, content) VALUES
 ('Kafka', '{"name":"Kafka","type":"org.apache.eagle.alert.engine.publisher.impl.AlertKafkaPublisher","description":null,"fields":[{"name":"kafka_broker","value":"sandbox.hortonworks.com:6667"},{"name":"topic"}]}'),
 ('Email', '{"name":"Email","type":"org.apache.eagle.alert.engine.publisher.impl.AlertEmailPublisher","description":null,"fields":[{"name":"subject"},{"name":"sender"}, {"name":"recipients"}]}'),
 ('Slack', '{"name":"Slack","type":"org.apache.eagle.alert.engine.publisher.impl.AlertSlackPublisher","description":null,"fields":[{"name":"token"},{"name":"channels"}, {"name":"severitys"}, {"name":"urltemplate"}]}'),
-('Storage', '{"name":"Storage","type":"org.apache.eagle.alert.app.AlertEagleStorePlugin","description":null,"fields":[]}');
+('HbaseStorage', '{"name":"HbaseStorage","type":"org.apache.eagle.alert.app.AlertEagleStorePlugin","description":null,"fields":[]}'),
```
haoch: "HbaseStorage" -> "HBaseStorage"

PR #771, opened 2017-01-11
EAGLE-858 JPM support task attempt info

zombieJ:
In JPM job detail UI, support display attempt info of failure job.


haoch:
LGTM :+1:

PR #782, opened 2017-01-19
[EAGLE-845][IMPROVEMENT] Using aggregation query in backend instead of calculating count in frontend.

chitin:
[IMPROVEMENT] Using aggregation query in backend instead of calculating count in frontend.
- Using siteId as query argument.
- Using aggregation query in backend instead of calculating count in frontend.


haoch:
LGTM, thanks for the contribution @chitin

PR #783, opened 2017-01-19
[EAGLE-872]  Transform counter metric to rate metric

r7raul1984:
- Add transform bolt using counterToRateFunction in HadoopMetricMonitorApp's storm topology.

https://issues.apache.org/jira/browse/EAGLE-872


haoch:
@r7raul1984 would you convert *all* counter metric to rate?

r7raul1984:
@haoch  If metric name endswith "count" , it will be converted to rate.

haoch:
@r7raul1984 how do you think expose `CountMetricFilter`  as a parameter to choose which metric to convert as `count`, and the current hard-code behavior should only be some kind of default implementation?

    @FunctionInterface
    interface  CountMetricFilter extends Function<Map, Boolean>
   
    DefaultCountMetricFilter = (metric) -> metric["metric"].endsWith("*.count")

haoch:
LGTM

r7raul1984:
@haoch  Thanks. I will merge.

PR #787, opened 2017-01-20
[MINOR] Change default-rack to /default-rack

r7raul1984:
- Change default-rack to /default-rack


qingwen220:
Why do you need this change?

qingwen220:
At the initial design, we prefer to extract the 'exact rack info', for example,  '/rack-45' -> 'rack-45',  '/' will be removed.

r7raul1984:
@qingwen220  Because ClusterNodeAPITopologyRackResolver use "/default-rack" .I think it will issue some problem when  you complete EAGLE-876.

qingwen220:
It do not matter.  The node with a new rack info will replace the old one

PR #791, opened 2017-01-24
[EAGLE-868] Refactor metric chart render process to improve user experience

r7raul1984:
- show "data loading icon" for each chart one by one
- load metric data query for each chart
- reader metric data into metric chart when response returned and remove "data loading icon" one by one

https://issues.apache.org/jira/browse/EAGLE-868


r7raul1984:
I will modify some typo in commit message,so I close the pr.

PR #794, opened 2017-02-07
Eagle-884 JPM support queue trend view

zombieJ:
* Add trend chart
* support queue drill down
* Capacity base line


qingwen220:
look good to me

PR #798, opened 2017-02-09
[EAGLE-864] Make sure "HBase" nav active in sidebar when visiting through URL

r7raul1984:
- When active submenu  then it's parent menu will active also.

https://issues.apache.org/jira/browse/EAGLE-864

### DIFF
--- a/eagle-server/src/main/webapp/app/dev/public/js/app.js
+++ b/eagle-server/src/main/webapp/app/dev/public/js/app.js
@@ -324,6 +324,22 @@ var app = {};
 			});
 
 			// ================================ Function ================================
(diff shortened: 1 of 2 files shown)


eagle-server/src/main/webapp/app/dev/public/js/app.js
```diff
@@ -324,6 +324,22 @@ var app = {};
 			});
 
 			// ================================ Function ================================
+			// Get parent side bar navigation item class by submenu
+			$scope.getPNavClass = function (subportals) {
```
zombieJ: `getPageNavClass` or `getPortalNavClass`? No need abbr.

PR #799, opened 2017-02-10
[EAGLE-890] Add Job Scheduling Framework and HadoopHealthCheckApp

haoch:
* Add Job Scheduling Framework based on quartz scheduling to support clustering job scheduling and HA 
* Add HadoopHealthCheckApp to firstly support HBase Health Monitor

https://issues.apache.org/jira/browse/EAGLE-890


jhsenjaliya:
@haoch , i think this would be very good feature to have in 0.5, what do you think?

PR #800, opened 2017-02-14
[EAGLE-892] fixng NPE on ResourceFetcher

jhsenjaliya:
MRRunningJobFetchSpout uses these functions which if returned Null, the follow up log line will throw NPE. ( as shown below )
```
apps = resourceFetcher.getResource(Constants.ResourceType.RUNNING_MR_JOB);
LOG.info("get {} apps from resource manager", apps.size());
```


r7raul1984:
LGTM

PR #804, opened 2017-02-15
[EAGLE-897]HBase JMX Widget Url is wrongly using siteName instead of siteId

chitin:
[EAGLE-897]HBase JMX Widget Url is wrongly using siteName instead of siteId
- Use siteId instead of siteName in hyperlink.

https://issues.apache.org/jira/browse/EAGLE-897


haoch:
LGTM

PR #811, opened 2017-02-16
[EAGLE-870] adding required packages into storm-jar

jhsenjaliya:
without these required packages, it throws class not found exception, more detail is in the ticket.


haoch:
LGTM

PR #816, opened 2017-02-17
EAGLE-910 support merge portal

zombieJ:
UI merge side bar item if same name


haoch:
Will fix the UT exception in #818

haoch:
Build failed because #818 which has already been fixed now, so merge this PR.

PR #825, opened 2017-02-20
[EAGLE-900] Only show left navigation item (HDFS, HBase) when we install the JMX application

chitin:
[EAGLE-900] Only show left navigation item (HDFS, HBase) when we install the JMX application
- Refactor eagle-hadoop-metric to single app for every service, for example, there is a eagle-hadoop-metric before, now eagle-hadoop-metric is split into eagle-hadoop-metric, eagle-hdfs-metric and eagle-hbase-metric, we can see HBase navigation item only if we install eagle-hbase-metric.

https://issues.apache.org/jira/browse/EAGLE-900


haoch:
LGTM

PR #834, opened 2017-02-22
EAGLE-902 Add JDBC data storage dependency into eagle server module

DadanielZ:
- add jdbc storage type dependency into eagle-service-base module
- port the fix for mysql table creation failure : [EAGLE-273](https://issues.apache.org/jira/browse/EAGLE-273)


haoch:
LGTM

PR #841, opened 2017-02-24
EAGLE-926: Alert engine fails to support more than one stream consumption from an application

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-926


wujinhu:
LGTM

haoch:
LGTM

PR #843, opened 2017-02-24
EAGLE-929 UI Policy definition stream group use group name

zombieJ:
Use group name instead of app name by stream group


qingwen220:
LGTM

PR #845, opened 2017-02-28
EAGLE-930 UI Notification support

zombieJ:
Support alert auto detect and popup notification of current eagle application.


haoch:
LGTM

PR #849, opened 2017-03-01
[EAGLE-934] eagle 0.5 assembly update

jhsenjaliya:


jhsenjaliya:
@haoch, can u please help reviewing this ?

eagle-assembly/src/main/conf/eagle.conf
```diff
@@ -75,7 +77,8 @@ metadata {
     username = "root"
     password = ""
     driverClassName = com.mysql.jdbc.Driver
-    url = "jdbc:mysql://server.eagle.apache.org:3306/eagle"
+    database = "eagle"
+    connection = "jdbc:mysql://server.eagle.apache.org:3306/eagle?user=root"
```
haoch: Is the `user=root` duplicated with `username` field?

eagle-server-assembly/src/main/bin/eagle-server.sh
```diff
@@ -1,136 +0,0 @@
-#!/bin/bash
```
haoch: Do you delete `eagle-env.sh` and `eagle-server.sh`?

haoch:
@jhsenjaliya The shell scripts seems no correct

eagle-assembly/src/main/conf/eagle.conf
jhsenjaliya: ya, i noticed that too, but kept whatever was there, let me remove it from connection

haoch:
LGTM

PR #850, opened 2017-03-02
EAGLE-941 reset timezone of time parse

zombieJ:
When user set backend timezone. Timerange picker will auth time failed


wujinhu:
LGTM

PR #864, opened 2017-03-10
[EAGLE-947] Publishers with same policy but different schema could produce duplicate alerts

garrettlish:
Assume that we have policy1 which have 2 kinds of output streams, one is stream1 and another is stream2. If publisher1 is configured for policy1 and stream1, and publisher2 is configured for policy1 and stream2, current code will produce 2 alerts for either stream1 or stream2.


RalphSu:
@wujinhu @haoch 
Please help review this one.

PR #872, opened 2017-03-14
Eagle-949 Support job suggestion

zombieJ:
JPM:
* Success job display job suggestions
* Failure Job display task attempt & detail

APP:
* UI support dialog window


qingwen220:
LGTM

PR #880, opened 2017-03-16
[MINOR] fix system-metric-collector collect disk info bug

koone:
`df -k` result line may be folded into two lines if filesystem is to long , add `P` can avoid this problem


haoch:
LGTM

PR #882, opened 2017-03-17
EAGLE-963 list & detail page use same durationTime

zombieJ:
job list page using time duration calculation. It's sync with detail page by durationTime now.


r7raul1984:
LGTM

PR #892, opened 2017-03-28
[MINOR] Fix Topology Check system config thrown NullPointerException bug

koone:
When 'dataSourceConfig.system.enabled' is false, NullPointerException will thrown out


qingwen220:
LGTM

baibaichen:
@r7raul1984, Please apply following rules for PRs from yhder

1. **For any bug fix, if there is no UT please reject PR.** 
2. **For any new feature of java end, if there is no UT please reject PR.**

PR #896, opened 2017-03-30
[EAGLE-981] GC overhead limit exceeded

asdf2014:
Travis throw `GC overhead limit exceeded` exception because the default `JVM` option of maven limit the resource of Travis.

https://issues.apache.org/jira/browse/EAGLE-981

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,3 @@ script:
   - mvn clean test -B
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


jhsenjaliya:
@asdf2014 , looks like this change was included in earlier PR already, please close this PR.

asdf2014:
@jhsenjaliya Okey.

PR #902, opened 2017-04-03
[EAGLE-985] Upgrade hbase dependency to 1.2.5

tedyu:
Upgrade hbase dependency to 1.2.5

### DIFF
--- a/eagle-core/eagle-embed/eagle-embed-hbase/src/main/java/org/apache/eagle/service/hbase/EmbeddedHbase.java
+++ b/eagle-core/eagle-embed/eagle-embed-hbase/src/main/java/org/apache/eagle/service/hbase/EmbeddedHbase.java
@@ -20,6 +20,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HBaseTestingUtility;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.MiniHBaseCluster;
+import org.apache.hadoop.hbase.TableName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -117,7 +118,7 @@ public class EmbeddedHbase {
 
     public void createTable(String tableName, String cf) {
         try {
-            util.createTable(tableName, cf);
+            util.createTable(TableName.valueOf(tableName), cf);
         } catch (IOException ex) {
             LOG.warn("Create table failed, probably table already existed, table name: " + tableName);
         }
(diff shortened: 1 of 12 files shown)


jhsenjaliya:
+1, LGTM, I will be testing this on my dev environment as well sometime tomorrow.

qingwen220:
@jhsenjaliya @tedyu  After this upgrade, does we still support HBase 0.98?

jhsenjaliya:
Hi @qingwen220 no, since there are some function signature changes here, I dont think it will be compatible. ( without those changes, it is compatible though) so the question really comes down to whether we want to support 0.98 till 0.5 release then this can be pushed to 0.6 when majority of the tech stack is going to change ( storm, kafka, etc..), comment?

qingwen220:
Yes. We expect after 0.5, Eagle can support Kafka 0.10 & storm 0.10. 

There is a concern. We have a coprocessor module in Eagle for HBase 0.98. I am not sure if we could support this function after upgrade

wujinhu:
Should we have multiple version support for HBase(Storm/Kafka etc) so that users can still use old versions if they cannot switch to new version quickly?
I think it will be better if we have some abstractions first instead of doing like this.

r7raul1984:
@wujinhu  +1

r7raul1984:
In my opition ,we should do more test before upgrade to hbase 1.2.5 rather than just solve compile problem .

jhsenjaliya:
master works fine with hbase 1.2.2
I am planning to test this patch with hbase 1.2.5 tonight

baibaichen:
hi @tedyu 

Is the new API you used  supported in 1.0.x?

tedyu:
Yes. But hbase 1.0 wouldn't have any more new release.

tedyu:
@jhsenjaliya 
Did you find any issue during testing ?

jhsenjaliya:
@tedyu I pulled your PR and used ur code to run my eagle, I am running alert engine app + hdfs audit log app and it successfully created the alert i was expecting. so this PR is working fine. but As i mentioned earlier, I was using hbase 1.2.2 and also tested hbase 1.2.5 without any code change to eagle and it was working fine as well. so probably these changes are not mandatory but good to follow newer hbase functions.

jhsenjaliya:
@qingwen220 , @r7raul1984 looks like we can accommodate hbase 1.2.5 with or without this PR changes but I havent tested with hbase coprocessor enabled. so if someone can test current eagle code with hbase 1.2.5 with coprocessor enabled, we can make decision on whether we can include hbase upgrade in eagle 0.5 or later.

PR #909, opened 2017-04-11
EAGLE-999: JobConfigSerDeser fails to serialize/deserialize data with long string 

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-999


wujinhu:
LGTM

PR #913, opened 2017-04-14
[EAGLE-1001] Compare Chart of JPM detail isn't sorted

r7raul1984:
- sort job list by startTime 

https://issues.apache.org/jira/browse/EAGLE-1001


r7raul1984:
@zombieJ  For now as a workaround, I will sort joblist by startTime in js engine. 
@qingwen220  In the future,I think GenericEntityService should sort by timestamp by default like MRJobExecutionResource does. And GenericEntityService should support sort by custom defined field.

Correct me if I'm wrong.

r7raul1984:
@qingwen220  I create a jira https://issues.apache.org/jira/browse/EAGLE-1003 for GenericEntityService  improvement.

zombieJ:
LGTM

PR #920, opened 2017-04-18
[EAGLE-1009] Fix `return` inside `finally` block may result in losing exception

asdf2014:
`return` inside `finally` block will result in losing exception:

* If the finally block completes abruptly for reason S, then the try statement completes abruptly for reason S (and the throw of value V is discarded and forgotten).
* If the finally block completes abruptly for reason S, then the try statement completes abruptly for reason S (and reason R is discarded).

reference:
http://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.2
https://issues.apache.org/jira/secure/attachment/12863778/FinallyTest.java

(https://issues.apache.org/jira/browse/EAGLE-1009)


qingwen220:
I don't think there are exceptions lost in the code you modified.  All these try statements are followed by catch statement which doesn't throw a new exception anymore. 

I agree with your idea that DON'T return in finally block

asdf2014:
@qingwen220 Hi, qingwen220. You are right. This situation will not lose any exception because all exception from `try` statement will be catched in `catch` block and it just print the exception message by `slf4j` in here.

PR #925, opened 2017-04-19
[EAGLE-1012] Some language level problems in `eagle-core` module

asdf2014:
Some language level problems
* Spell
* Duplicated
* Lambda
* Collection
* String
* instanceof
* Complex Method
* Exception
* AutoClose
* Synchronized

Details see [attachment files](https://issues.apache.org/jira/secure/attachment/12863947/Eagle%20Language%20Level%20Problems.txt).

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert-service/src/main/java/org/apache/eagle/service/alert/resolver/AttributeResolverFactory.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert-service/src/main/java/org/apache/eagle/service/alert/resolver/AttributeResolverFactory.java
@@ -16,8 +16,8 @@
  */
 package org.apache.eagle.service.alert.resolver;
 
-import org.apache.eagle.metadata.service.ApplicationEntityService;
 import com.typesafe.config.Config;
+import org.apache.eagle.metadata.service.ApplicationEntityService;
 
 import java.util.Collections;
 import java.util.HashMap;
@@ -43,8 +43,6 @@ public final class AttributeResolverFactory {
                 fieldResolvableCache.put(fieldResolverName, instance);
             } catch (ClassNotFoundException e) {
                 throw new AttributeResolveException("Attribute Resolver in type of " + fieldResolverName + " is not found", e);
-            } catch (InstantiationException | IllegalAccessException e) {
-                throw new AttributeResolveException(e);
             } catch (Exception ex) {
                 throw new AttributeResolveException(ex);
             }
--- a/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertKafkaPublisher.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertKafkaPublisher.java
@@ -18,13 +18,7 @@
 
 package org.apache.eagle.alert.engine.publisher.impl;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-
+import com.typesafe.config.Config;
 import org.apache.eagle.alert.engine.coordinator.Publishment;
 import org.apache.eagle.alert.engine.coordinator.PublishmentType;
 import org.apache.eagle.alert.engine.model.AlertStreamEvent;
@@ -37,7 +31,11 @@ import org.apache.kafka.clients.producer.RecordMetadata;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.typesafe.config.Config;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 public class AlertKafkaPublisher extends AbstractPublishPlugin implements AlertPublishPluginProvider {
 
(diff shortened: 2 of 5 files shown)


asdf2014:
If these improvements are necessary, we should step by step merge to avoid too much code conflict. So, creating multiple sub `PRs` for `EAGLE-1012` will be more suitable for this situation. What do you think?  @jhsenjaliya @qingwen220

eagle-core/eagle-alert-parent/eagle-alert-service/src/main/java/org/apache/eagle/service/alert/resolver/AttributeResolverFactory.java
```diff
@@ -43,8 +43,6 @@ public static AttributeResolvable getAttributeResolver(String fieldResolverName,
                 fieldResolvableCache.put(fieldResolverName, instance);
             } catch (ClassNotFoundException e) {
                 throw new AttributeResolveException("Attribute Resolver in type of " + fieldResolverName + " is not found", e);
-            } catch (InstantiationException | IllegalAccessException e) {
```
jhsenjaliya: instead of removing this, it can be used to describe the error, IDE is suggesting to collapse the catch block may be because it has identical error msg (throw new AttributeResolveException(e)).

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertEagleStorePlugin.java
```diff
@@ -62,10 +60,9 @@ public void onAlert(AlertStreamEvent event) throws Exception {
         if (eventList == null || eventList.isEmpty()) {
             return;
         }
-        List<AlertPublishEvent> alertEvents = new ArrayList<>();
-        for (AlertStreamEvent e : eventList) {
-            alertEvents.add(AlertPublishEvent.createAlertPublishEvent(e));
-        }
+        List<AlertPublishEvent> alertEvents = eventList.stream().map(
```
jhsenjaliya: i feel its fancy to use newer java features, but old syntax is more readable, may be its just a coding preference

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertPublisherImpl.java
```diff
@@ -199,7 +194,7 @@ private void closePlugins(List<AlertPublishPlugin> toBeClosed) {
             try {
                 p.close();
             } catch (Exception e) {
-                LOG.error(String.format("Error when close publish plugin {}!", p.getClass().getCanonicalName()), e);
+                LOG.error(String.format("Error when close publish plugin %s!", p.getClass().getCanonicalName()), e);
```
jhsenjaliya: people are using this interchangeably

jhsenjaliya:
@asdf2014 , i havent gone through all the changes yet, but looks like these are what some particular IDE would recommend, some are real improvements as well. lets check this out after 0.5, Thanks for the efforts !

eagle-core/eagle-alert-parent/eagle-alert-service/src/main/java/org/apache/eagle/service/alert/resolver/AttributeResolverFactory.java
asdf2014: I see. I will roll it back.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertEagleStorePlugin.java
asdf2014: I think it will bring better maintainability and [performance](http://blog.takipi.com/benchmark-how-java-8-lambdas-and-streams-can-make-your-code-5-times-slower/).

asdf2014:
@jhsenjaliya Okay, i see. Thx for your `code review`.

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertPublisherImpl.java
asdf2014: `String.format` should use `%s` rather than `{}`

```java
@Test
public void testFormat() {
    assertEquals("a{}c", String.format("a{}c", "b"));
    assertEquals("abc", String.format("a%sc", "b"));
}
```

eagle-core/eagle-alert-parent/eagle-alert/alert-engine/src/main/java/org/apache/eagle/alert/engine/publisher/impl/AlertEagleStorePlugin.java
asdf2014: Maybe you are right, i will roll it back. :+1:

asdf2014: I have created a non-professional benchmark [test](https://github.com/asdf2014/yuzhouwan/blob/master/yuzhouwan-hacker/src/main/java/com/yuzhouwan/hacker/lambda/StreamPerformance.java#L19). After adding `-Xmx2G -Xms2G -Xmn1G -XX:+AlwaysPreTouch` params for reducing the impact from jvm resources, then found the performance gap is not very large, and the performance of `lambda stream` is better than `simple for loop` after the data becomes larger.

jhsenjaliya:
@asdf2014 Thanks again, lets work on this after 0.5

asdf2014:
@jhsenjaliya You are welcome. Sorry, i just note it. Let’s finish this after `v0.5.0` together.

PR #936, opened 2017-05-17
[EAGLE-1025] bumping version to 0.5.1 for master

jhsenjaliya:
[EAGLE-1025] bumping version to 0.6 for master


coheigea:
The patch doesn't apply any more, as it assumes the current version is 0.5.0-SNAPSHOT.

PR #939, opened 2017-05-26
EAGLE-1029: fix bug in generateGroupbyMonitorMetadata

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-1029


qingwen220:
This bug can be produced as follows

1. create policies with stream1 to use up all the working slot of alert engine topology1
2. setup another alert engine topology topology2
3. create a policy with stream1 and trigger the scheduler to schedule
4. check schedule result by GET rest/metadata/schedulestates, and compare groupSpecs & alertSpecs

RalphSu:
Make sense.

@qingwen220  can you please add a unit test case that reproduce this issue and make sure we fix it?

qingwen220:
@RalphSu  I have added a unit test. This test verifies the correctness of assigning of a policy to a new alert engine topology.

PR #942, opened 2017-06-05
Eagle-1032 Add policy duplication settings in the policy definition page

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1032

Add policy duplication settings in the policy definition page.
1. FieldName: dedupIntervalMin Type: string (text) - default: 30
e.g., 1, 2, 10
2. FieldName: dedupFields Type: array (checkbox) - default: all column checked
Description: show all stream columns of STRING type
e.g. ["site", "component"]


qingwen220:
LGTM

PR #943, opened 2017-06-07
[MINOR] Fxi sprk/egle issues in JPM module

asdf2014:
Fxi sprk issues in JPM module


qingwen220:
@asdf2014  I do not think these are big issues. 

On the other hand, this changes will have a big impact for users who have already installed these features. (old data are missing).

asdf2014:
@qingwen220 Yep, you are right. This will destroy backward compatibility. But, for people with obsessive compulsive disorder it still be a huge problem. Moreover, if you do not modify it now, then want to modify in future, it will become more difficult.

jhsenjaliya:
@asdf2014 i know how it feels.
but as @qingwen220 pointed out, it will be too much work to make this compatible for people already running on branch 0.5 
how about we take up this on next version 0.6 ?

asdf2014:
@jhsenjaliya Okay, i see :D

PR #944, opened 2017-06-07
EAGLE-1038: Support alertDuplication customization for each policy

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-1038

* support duplication check for each outputStream of a policy
* compatible with the duplication check in old versions (check in a publisher)

### DIFF
--- a/eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/model/AlertStreamEvent.java
+++ b/eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/model/AlertStreamEvent.java
@@ -39,6 +39,7 @@ public class AlertStreamEvent extends StreamEvent {
     private long createdTime;
     private String category;
     private AlertSeverity severity = AlertSeverity.WARNING;
+    private boolean duplicationChecked = false;
 
     // ----------------------
     // Lazy Alert Fields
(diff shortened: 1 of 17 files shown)


eagle-core/eagle-alert-parent/eagle-alert/alert-common/src/main/java/org/apache/eagle/alert/engine/model/AlertStreamEvent.java
```diff
@@ -39,6 +39,7 @@
     private long createdTime;
     private String category;
     private AlertSeverity severity = AlertSeverity.WARNING;
+    private boolean duplicationChecked = false;
```
asdf2014: Just using `private boolean duplicationChecked;` is fine, because the default value of `boolean` in jdk is false.

qingwen220: what if the default value changes in different JDK versions?

asdf2014: I think it is impossible to change it, because there is no other better way to [design](https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5) it and redesignning it will bring too much impact ...

PR #947, opened 2017-06-13
EAGLE-1041: Support policy processing pipeline

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-1041

Two updates:
* if an inputStream is an intermediate stream (defined by select clause statement), then remove it from inputStream list and outputStream list
* if an inputStream is an intermediate stream, remove its PartitionSpec


zombieJ:
Remove `console` & LGTM

PR #948, opened 2017-06-14
EAGLE-1042 support preview chart title

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1042

For metric preview page. There should be display the title of the chart for quick snapshot


qingwen220:
LGTM

PR #949, opened 2017-06-14
EAGLE-1039 recheck when policy init and keep origin setting

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1039

If a user wants to edit an existing policy, alert de-duplication info should be loaded from the database when he/she enters to that page.


qingwen220:
LGTM

PR #950, opened 2017-06-14
EAGLE-1043 display date when series cross the date

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1043

Current chart don't provide date info which is not user friendly


qingwen220:
LGTM

PR #953, opened 2017-06-20
EAGLE-1049 Support metric filter in the metric preview page 

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1049

Support metric filter in the metric preview page


qingwen220:
LGTM

PR #954, opened 2017-06-20
EAGLE-1047 View all alert publishers on Eagle UI

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1047

Users can view all the publisher configurations


qingwen220:
LGTM

PR #955, opened 2017-06-20
EAGLE-1048 Delete an alert publisher on Eagle UI

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1048

Administer can delete  an alert publisher on Eagle UI

Request: DELETE /publishments/\{name\}

Response: 
```
OpResult {
    public int code = 200;   // 200 = SUCCESS 
    public String message = "";
}
```


qingwen220:
LGTM

PR #956, opened 2017-06-21
EAGLE-1054 Publisher delete don't close the dialog

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1054

When user delete the publisher. The dialog not closed.


qingwen220:
LGTM

PR #958, opened 2017-06-21
EAGLE-1055: update policy proto APIs

qingwen220:
https://issues.apache.org/jira/browse/EAGLE-1055


zombieJ:
Greate. Let's merge into master

PR #963, opened 2017-07-11
Eagle-1046 Eagle supports policies import to a new site from a policy prototype

zombieJ:
https://issues.apache.org/jira/browse/EAGLE-1046

add prototype management page
support policy create with proto


qingwen220:
+1. A very useful feature for policy management on multiple sites.

wujinhu:
LGTM

PR #991, opened 2018-04-21
EAGLE-1088: Change the inconsistent method name 'find'.

BruceKuiLiu:
Change the method name 'find' to 'isMatched'.
The method is named "find", but the method does not find anything, just returns a boolean-variable 'this.matched'. Thus, rename method as "isMatched" should be more clear than "find".


jhsenjaliya:
@BruceKuiLiu, I believe this is consistent with Java.Pattern class method where find will return true/false based on pattern match. 
I would let @yonzhang comment further on this. 
Thanks for the PR though.

PR #992, opened 2018-05-20
[EAGLE-1090] - update for compatibiliy to python3

jhsenjaliya:
- used requests instead of urllib
- updated print statements
- used input instead of raw_input


jhsenjaliya:
why this was closed and not merged?

